PR #25, opened 2018-04-30
Fix index out of range issue

bgeesaman:
Before:
```
NAME                         	REVISION	UPDATED                 	STATUS  	CHART                             	NAMESPACE  
test-ingress                 	3       	Mon Apr 30 13:43:05 2018	DEPLOYED	k8s-ingress-0.1.1                 	kube-systemkube-system
```

When `helm_helpers.go:82` goes to pick up the 10th field, it doesn't exist because it's a part of field 9.  Adding a space ensures the columns don't run together.

After:
```
NAME                         	REVISION	UPDATED                 	STATUS  	CHART                             	NAMESPACE  
test-ingress                 	3       	Mon Apr 30 13:43:05 2018	DEPLOYED	k8s-ingress-0.1.1                 	kube-system kube-system
```

### DIFF
--- a/helm_helpers.go
+++ b/helm_helpers.go
@@ -47,7 +47,7 @@ func getTillerReleases(tillerNS string) string {
 	lines := strings.Split(result, "\n")
 	for i, l := range lines {
 		if l != "" && !strings.HasPrefix(l, "NAME") && !strings.HasSuffix(l, "NAMESPACE") {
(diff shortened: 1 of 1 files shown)


sami-alajrami:
good catch @bgeesaman , actually the space is not the problem .. it is in the for loop in buildState() .. It has wrong iterations config which I forgot to change earlier. Will fix that.

Issue #35, opened 2018-05-25
support a bit better variable munging


tibers:
My golang isn't good enough to pull this off but I see that [variable substitution is limited to the names of variables but no string interpolation](https://github.com/Praqma/helmsman/blob/dec253e7268fa76406763bddd8d60ec9f9b431d8/state.go#L158-L166).

This means that a desirable behavior like `appName = "MySimpleApp.$fqdn"` can't happen, even though it seems like a natural pattern since `appName = "$HOSTNAME"` can because the code finds the leading `$` and looks up the environment variable.

I'd like to propose the normal pattern of `STRING-$VARIABLENAME-STRING` where the code would understand that any patterns of letters after a `$` is a variable name.

nicolaiskogheim:
To remove the requirement to not have alphanums after a variable name in shell scripting (sh, bash, etc.), an alternate syntax is used: `${HOSTNAME}`, which means you could have `STRING${VARIABLENAME}STRING` if you so desire. Might be worth considering to allow the option of curly braces, or even just straight out require those braces.

tibers:
I think the project is dead. We moved onto helmfile: https://github.com/roboll/helmfile

nicolaiskogheim:
Thanks for the heads-up and for the alternative!

sami-alajrami:
@tibers @nicolaiskogheim apologies for the late response as I was away from this for awhile .. and thanks for this fix :)

PR #36, opened 2018-05-25
trying to enable environmental parsing across the board in helmsman

tibers:
I am attempting to [fix this issue](https://github.com/Praqma/helmsman/issues/35). I am a total go novice and I need another set of eyes.


state.go
```diff
@@ -159,8 +159,8 @@ func (s state) validate() (bool, string) {
 // if the env variable is empty or unset, an empty string is returned
 // if the string does not start with '$', it is returned as is.
 func subsituteEnv(name string) string {
```
davidarcher: Looks like this function essentially is the same thing as os.ExpandEnv so maybe just drop it completely?

tibers: I was going to make a "A WILD $THING APPEARS" joke but I couldn't find one SFW... I updated the PR because why not?

tibers:
OK, fixed all the whitespace. Now there won't be some files with two spaces and some with five.

jonbesga:
I assume this means that no dollar values can't be used then?

I have a password like `1111$2222` and it gets translated into the chart as 1111 because of this PR

sami-alajrami:
@jonbesga you can escape the $ like so: `1111$$2222` or alternatively, if you are using v1.10.x (and don't need variable substitution anywhere else), you can use  the `--no-env-subst` flag to disable variable substitution everywhere.

PR #67, opened 2018-08-30
Update config.yml

luisdavim:
missed a few spots, maybe adding a `Makefile` could simplify this process.


sami-alajrami:
@luisdavim sure ..  I am not that familiar with `make` but feel free to try it out :)

PR #69, opened 2018-09-02
adding a Makefile

luisdavim:


luisdavim:
interesting, I'll look into that. but it looks like `gcc` is missing

luisdavim:
I found the issue, I had to add `--always` so git describe falls back to an abbreviated hash if it can't find any suitable tags.

luisdavim:
Also fixed the `make test`, the `-race` flag doesn't work on Alpine... :(

PR #70, opened 2018-09-03
Adding new option to reverse the priorities on delete

luisdavim:


luisdavim:
@bergerx because otherwize I would need to create a new structure for the settings, it mapped to `map[string]string`

bergerx:
Oh I see, it's coming from
https://github.com/Praqma/helmsman/blob/71e1165ce72d87536868a3134d291bb5bfe80e46/state.go#L27

Extending the state.go to address this doesn't seem hard, but could create
some unexpected overhead on this PR since you have to touch multiple
unrelated places which uses the settings map.

Sorry if I caused unnecessary overhead.

luisdavim:
Yes, no worries, I'll look into it tomorrow.

sami-alajrami:
boolean would be more consistent with the other options. A simple solution to avoid the overhead of modifying the settings map is to parse the string into boolean .. something like: `strconv.ParseBool()` which also requires validating/handling of none boolean string values.

luisdavim:
I've created a new struct for the settings, please revire @bergerx and @sami-alajrami

state.go
```diff
@@ -20,11 +20,23 @@ type namespace struct {
 	ClientKey            string `yaml:"clientKey"`
 }
 
+// config type represents the settings fields
+type config struct {
```
bergerx: Why the struct name is `config` and not `settings`?

luisdavim: It was to make sure it wasn't going to conflict with any other existing variable. I can change it but, it is that important?

luisdavim:
Thanks for reviewing so quickly @bergerx .

state.go
bergerx: Not at all, that fine as is.

sami-alajrami:
LGTM. Thanks @luisdavim !

PR #74, opened 2018-09-09
adding helm diff to the decision maker

luisdavim:
Here's an example output:
```
$ helmsman -f helmsman-common.yaml -f dev/helmsman-dev.yaml
 _          _
| |        | |
| |__   ___| |_ __ ___  ___ _ __ ___   __ _ _ __
| '_ \ / _ \ | '_ ` _ \/ __| '_ ` _ \ / _` | '_ \
| | | |  __/ | | | | | \__ \ | | | | | (_| | | | |
|_| |_|\___|_|_| |_| |_|___/_| |_| |_|\__,_|_| |_|version: v1.5.0
A Helm-Charts-as-Code tool.


2018/09/09 13:31:16 INFO: Parsed YAML [[ helmsman-common.yaml ]] successfully and found [ 0 ] apps.
2018/09/09 13:31:16 INFO: Parsed YAML [[ dev/helmsman-dev.yaml ]] successfully and found [ 1 ] apps.
2018/09/09 13:31:16 INFO: namespace validation -- Tiller is NOT desired to be deployed in namespace [ kube-system ].
2018/09/09 13:31:16 INFO: namespace validation -- Tiller is desired to be deployed WITHOUT TLS in namespace [ dev ].
2018/09/09 13:31:19 WARN: I could not create namespace [dev ]. It already exists. I am skipping this.
2018/09/09 13:31:20 WARN: I could not create namespace [kube-system ]. It already exists. I am skipping this.
2018/09/09 13:31:20 INFO: deploying Tiller in namespace [ dev ].
2018/09/09 13:31:28 INFO: checking what I need to do for your charts ...
2018/09/09 13:31:28 INFO: mapping the current helm state ...
dev, artifactory-artifactory, Ingress (extensions) has been added:
-
+ # Source: artifactory/templates/ingress.yaml
+ apiVersion: extensions/v1beta1
+ kind: Ingress
+ metadata:
+   name: artifactory-artifactory
+   labels:
+     app: artifactory
+     chart: artifactory-7.0.6
+     release: artifactory
+     heritage: Tiller
+   annotations:
+ spec:
+   rules:
+     - host: artifactory.domain.example
+       http:
+         paths:
+           - path: /
+             backend:
+               serviceName: artifactory-artifactory
+               servicePort: 8081

2018/09/09 13:31:34 INFO: checking if any Helmsman managed releases are no longer tracked by your desired state ...
2018/09/09 13:31:35 INFO: no untracked releases found.
[...]

PR #76, opened 2018-09-10
don't run as root

luisdavim:
This should fix the issues caused by #74.
Sorry for overlooking that.


sami-alajrami:
awesome @luisdavim , I have tested it and it works .. thanks!

Issue #77, opened 2018-09-12
better validation and review of "plans"


rsanders:
We have a 2 part helmsman CI run -- first stage runs without `--apply` to validate everything. If that succeeds, we run again with `--apply`.

Similarly, CI for branches in the repo will just run the 1st stage. A developer then inspects the output and decides whether to merge.

Unfortunately, running without `--apply` isn't `terraform plan` or `helm --dry-run`, where you can be relatively sure that simple validation errors will be caught in the plan/dry-run phase.  Problems we've hit when doing `--apply` that weren't caught when run without `--apply`:

- Missing values file / bad values file path
- Failure to render Helm chart due to YAML errors
- Failure to install/upgrade Helm chart due to K8S API validation errors
- Missing helm chart or non-matching helm chart version (IIRC)

It's also very difficult to read the helmsman output to understand what the effect of a run will be.

It would be incredibly useful to have a `--dry-run` or `helmsman plan` function that did as much validation as possible, and showed a simple, easy-to-read plan of decisions & actions that would result from running with `--apply`.

See also #60

luisdavim:
We've recently introduced helm-diff to the decision making logic, that should address some of the problems you've pointed out. Have you checked that out?

luisdavim:
But I still see room for a `--dry-run` option in helmsman to address the rest of the issues you mentioned.

sami-alajrami:
I will work on the dry-run option tonight.

sami-alajrami:
@rsanders I have added the `--dry-run` which applies the helm `--dry-run --debug` options when executing commands from Helmsman. This catch problems like:
> - Failure to render Helm chart due to YAML errors

For issues like: 
> - Missing values file / bad values file path
> - Missing helm chart or non-matching helm chart version (IIRC)

These should be caught by existing validation, if you encounter these issues, then please let us know with an example to fix it :)

Fot this one:
> Failure to install/upgrade Helm chart due to K8S API validation errors

It is a bit tricky to discover it before helm renders your templates and send them to K8S API for applying them. It also varies based on K8S versions and so on. Even `terraform plan` does not do such validation.

sami-alajrami:
I have also made some initial effort on color coding Helmsman output to highlight the plan , error messages and Helm's output.

sami-alajrami:
The `--dry-run` option is shipped in v1.6.0. I am going to close this issue for now, but feel free to reopen if you are still experiencing issues :)

Issue #60, opened 2018-08-20
Add option to show chart render


txomon:
During development, it would be really usefull to see how all the templates have been rendered for debugging purposes. Something similar to `helm template` subcommand would be really helpful.

sami-alajrami:
we could add `dry-run` option to use helm's dry-run option when installing/upgrading releases. That will render your templates and output it. Would that be good?

txomon:
@sami-alajrami anything with the functionality of giving such output is good

sami-alajrami:
The `--dry-run` option is shipped in v1.6.0. I am going to close this issue for now, but feel free to reopen if you are still experiencing issues :)

PR #81, opened 2018-09-16
adding dry-run option

sami-alajrami:


sami-alajrami:
this fixes #77 and #60

Issue #91, opened 2018-09-27
make build is failing


luisdavim:
Hi,

I pulled the latest code and got this when running `make build`:
```
$ make build
# _/Users/ldavim/workingCopies/code/github/PTC-Global/helmsman
./utils.go:181:15: too many arguments in call to aws.ReadFile
        have (string, string, string, bool)
        want (string, string, string)
./utils.go:186:15: too many arguments in call to gcs.ReadFile
        have (string, string, string, bool)
        want (string, string, string)
make: *** [build] Error 2
```

luisdavim:
it worked after I removed some of the cached dependencies from `~/go/src/` maybe we should start using something like https://golang.github.io/dep/docs/introduction.html to manage dependencies and their versions.

luisdavim:
https://github.com/blindpirate/report-of-build-tools-for-java-and-golang#golang

sami-alajrami:
@luisdavim I have added the dep files. 
As those two packages in the above issue are internal, changes to them will have to be pushed to github first before `dep ensure` can get the updated dependencies.

luisdavim:
we should update the makefile to use godep instead of fetching all the libs.

luisdavim:
https://github.com/Clever/dev-handbook/blob/master/golang/dep.md

luisdavim:
I can open a PR for that if you want.

luisdavim:
ok, introducing `dep` forces people that want to contribute to this project to clone it into a `GOPATH/src` directory, I can workaround it in the make file and keep things as they where without `dep` and anyone can clone this repo into any path and all `make` commands will just work, that will had some complexity to the `Makefile`.
If you don't think that's important, we can keep the make file simple and contributors will have to abide to the `GOPATH`.
I can open a `WIP` PR with the "complex" version for you to see what I mean....

PR #96, opened 2018-09-28
Updating the makefile to use dep to ensure the necessary dependencies

luisdavim:


luisdavim:
I thought that the build and tests in the CI where happening in the `/go/src/helmsman` but they are in `/home/helmsman/project`, so now this is failing because in that docker image we can't `cd` into `/tmp` but even if I remove all the extra logic to allow the build to work even if the project is not cloned int o a `GOPATH/src` directory this will fail because the project is not being cloned into a `GOPATH/src` directory....

luisdavim:
Is the CI tool using the `dockerfile` from `test_files`? If I run:

```
$ docker build -t helmsman-test test_files/
#....
Successfully built 720721c6ab5a
Successfully tagged helmsman-test:latest
$ docker run --rm -it helmsman-test sh
$ git init
#...
$ git remote add -t dep_make -f origin https://github.com/PTC-Global/helmsman.git
#...
git checkout dep_make
#...
make test
#...
PASS
coverage: 21.2% of statements
ok      helmsman        0.791s
exit
```

so why doesn't it work in the CI?

luisdavim:
@sami-alajrami can you check why this isn't running in the docker container created with `test_files/dockerfile`?

sami-alajrami:
@luisdavim if checking the code out under GOPATH is all what's needed, I have changed the working directory in circleci (where the source code will be checked out) to `/go/src/helmsman/`.

sami-alajrami:
the `helmsman-test` docker image is manually pushed and I see that you have added `dep` to it in your branch. The one used by the CI didn't have `dep` .. not sure if this would be the cause of the error you were getting. I have pushed a new image which contains `dep`.

Makefile
```diff
@@ -1,10 +1,29 @@
...
+
+GOPATH := $(realpath $(GOPATH))
+ifeq ($(strip $(GOPATH)),)
+  $(error GOPATH is not set and could not be automatically determined)
+endif
+
+SRCDIR := $(GOPATH)/src/
+
+ifeq ($(filter $(GOPATH)%,$(CURDIR)),)
+  GOPATH := $(shell mktemp -d "/tmp/dep.XXXXXXXX")
```
sami-alajrami: now that the source code is checkout under GOPATH, maybe you don't need to change GOPATH into /tmp anymore?

luisdavim: This is about what I wrote on #91 we can drop this and make the makefile simpler but that will force  people that want to contribute to clone this into a `GOPATH/src` folder. With this implementation the temp folder is only used if the project is cloned outside of `GOPATH`.

luisdavim:
Regarding the `helmsman-test` the problem was not just the missing `dep` package, the code was not being deployed under the workdir defined in the Dockerfile. As you can see in the example I posted before if you run the tests in that container the codes does go under `/go/src` it looks like it wasn't the case the an older version of the image but that the image was not being used.

luisdavim:
@sami-alajrami do you think we can get this merged? I think it's better to leave it as it is so it's backwards compatible for contributors that don't like keeping their forks under a folder in `GOPATH` and the tmp folder is only used in that case.

sami-alajrami:
thanks @luisdavim  .. yes, I will merge it and if it still fails in the CI we will get to that separately.

Issue #100, opened 2018-10-04
Keep helmsman from creating namespace


bobhenkel:
In our current setup we have a helm chart that creates namespaces to our platform teams spec that dev teams can use, it setups some namespace roles, resource quotas, and service account for the namespace to name a few. We put a super tiller/root tiller/god tiller whatever you want to call it in a tools namespace that the platform team creates on cluster creation.  Once that tiller is up we can install our helm chart namespace-creator to create new namespaces and install a tiller into the namespace being created so that the team that owns the namespace is tiller scoped to their namespace. With helmsman I'm running into the issue that it wants to create the namespace as soon as I mention it and if it doesn't exists, and in our case we don't want helmsman to manage our namespace out of band(not a helm chart install). Take this example toml.

```
[namespaces]
  [namespaces.bob-helmsman]
    protected = false
    useTiller = false
  [namespaces.kube-system]
    protected = true
    installTiller = false
  [namespaces.tools]
    protected = true
    useTiller = true

[helmRepos]
  stable = "https://kubernetes-charts.storage.googleapis.com"
  bob-repo =  "http://127.0.0.1:8080"


[apps]

  [apps.namespace-creator]
    namespace = "bob-helmsman" # maps to the namespace as defined in namespaces above
    enabled = true # change to false if you want to delete this app release [default = false]
    chart = "bob/namespace-creator" # changing the chart name means delete and recreate this release
    version = "0.1.9" # chart version
    ### Optional values below
    name = "namespace-creator" # should be unique across all apps which are managed by the same Tiller
    valuesFile = "" # leaving it empty uses the default chart values
    tillerNamespace = "tools" # which Tiller to use to deploy this release
    purge = true # will only be considered when there is a delete operation
    test = false # run the tests when this release is installed for the first time only
    protected = false
    priority= -100
    [apps.namespace-creator.set]
      "namespaceQuota.enabled" = "true"
      "namespaceQuota.requestsCpu" = "400m"
      "namespaceQuota.requestsMemory" = "1Gi"
      "namespaceQuota.limitsCpu" = "500m"
      "namespaceQuota.limitsMemory" = "1Gi"
      "podLimit.pods" = "100"
      "labels.environment" = "live"
      "labels.slackchannel" = "bob"
      "labels.team" = "bob"
```

The problem is helmsman will try to create the bob-helmsman namespace before my app install can run the helm install for namespace-creator. Is there any way to tell helmsman that hey here's a namespace, but don't attempt to create it?

[...]

Issue #65, opened 2018-08-27
Helmsman create namespaces and deploys Tiller without `--apply`


epierotto:
Helmsman is creating namespaces and deploying Tiller without `--apply`.
My expectation is that Helmsman should only apply changes if its run with `--apply`

```bash
kubectl version
Client Version: version.Info{Major:"1", Minor:"11", GitVersion:"v1.11.2", GitCommit:"bb9ffb1654d4a729bb4cec18ff088eacc153c239", GitTreeState:"clean", BuildDate:"2018-08-08T16:31:10Z", GoVersion:"go1.10.3", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"10", GitVersion:"v1.10.6", GitCommit:"a21fdbd78dde8f5447f5f6c331f7eb6f80bd684e", GitTreeState:"clean", BuildDate:"2018-07-26T10:04:08Z", GoVersion:"go1.9.3", Compiler:"gc", Platform:"linux/amd64"}
```
```bash
helm version
Client: &version.Version{SemVer:"v2.10.0", GitCommit:"9ad53aac42165a5fadc6c87be0dea6b115f93090", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.10.0", GitCommit:"9ad53aac42165a5fadc6c87be0dea6b115f93090", GitTreeState:"clean"}
```

**This is only happening if you have at least one app defined**
Output without apps defined:
```bash
helmsman -f myfile.yaml --debug
 _          _
| |        | |
| |__   ___| |_ __ ___  ___ _ __ ___   __ _ _ __
| '_ \ / _ \ | '_ ` _ \/ __| '_ ` _ \ / _` | '_ \
| | | |  __/ | | | | | \__ \ | | | | | (_| | | | |
|_| |_|\___|_|_| |_| |_|___/_| |_| |_|\__,_|_| |_|version: v1.4.0-rc
A Helm-Charts-as-Code tool.


2018/08/27 13:09:13 INFO: validating that helm is installed.
2018/08/27 13:09:13 INFO: validating that kubectl is installed.
2018/08/27 13:09:13 INFO: Parsed YAML [[ myfile.yaml ]] successfully and found [ 0 ] apps.
2018/08/27 13:09:13 INFO: namespace validation -- Tiller is desired to be deployed WITHOUT TLS in namespace [ test1 ].
[...]

PR #101, opened 2018-10-12
adding flag to disable namespace creation

luisdavim:
trying to address #100

### DIFF
--- a/helm_helpers.go
+++ b/helm_helpers.go
@@ -88,6 +88,11 @@ func getTillerReleases(tillerNS string) tillerReleases {
 
 	exitCode, result := cmd.exec(debug, verbose)
 	if exitCode != 0 {
+		if !apply {
+			log.Println("INFO: " + strings.Replace(result, "Error: ", "", 1))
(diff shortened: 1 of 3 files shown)


sami-alajrami:
this could be addressed in a more comprehensive approach by trying to address #65 .. it might be better to prevent creating anything in the cluster (namespaces, Tiller deploys, service accounts  etc) if the `--apply` is not used. I haven't checked in detail yet what impacts this would have on generating a plan .. what do you think @luisdavim ?

luisdavim:
makes sense, I'll look into that.

luisdavim:
@sami-alajrami I didn't had much time to work on this before but I think that with the changes that I've pushed today, this should address both #65 and #100 . can you take a look and let me know what you think?

I've kept the `--no-ns` flag anyway because users like @bobhenkel might still need it.

sami-alajrami:
thanks @luisdavim .. I will have a look today.

sami-alajrami:
this looks good .. thanks again @luisdavim :)

PR #102, opened 2018-10-16
Fixes diff error when wait flag is set.

SteveRuble:
The diff plugin errors if you pass the --wait flag, and I don't think --wait could mean anything in this context, so I removed it from the invocation.

I also updated the description of the command run by diffRelease so that it reflects what the command is doing.

### DIFF
--- a/decision_maker.go
+++ b/decision_maker.go
@@ -225,8 +225,8 @@ func diffRelease(r *release) string {
 
 	cmd := command{
 		Cmd:         "bash",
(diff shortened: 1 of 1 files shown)


hatemosphere:
@luisdavim any chances to get this in a new release any time soon? :)

sami-alajrami:
@hatemosphere  v1.6.3 is released with this fix .. docker images are being built now :)

sami-alajrami:
@hatemosphere v1.6.3 is released with this fix. docker images are being built now :)

PR #118, opened 2018-11-02
delete untracked releases

### LINKED ISSUE #117: does not delete untracked releases in namespaces that don't have a tiller installed
When only using on tiller instance (e.g. in kube-system) it's possible to deploy applications to other namespaces than `kube-system` but they're not removed when they are removed from the DSF. That is because helmsman only check for untracked releases in the namespace where tiller is deployed:

```
2018/11/01 14:56:48 INFO: checking if any Helmsman managed releases are no longer tracked by your desired state ...
2018/11/01 14:56:48 INFO: getting helm releases which are managed by Helmsman in namespace [[ kube-system ]].
2018/11/01 14:56:49 INFO: getting helm releases which are managed by Helmsman in namespace [[ kube-system ]].
2018/11/01 14:56:49 INFO: no untracked releases found.
```

Also, awesome work on this, and let me know if I can help with getting this fixed :)

despairblue:
fix #117

### DIFF
--- a/kube_helpers.go
+++ b/kube_helpers.go
@@ -353,7 +353,7 @@ func getHelmsmanReleases() map[string]map[string]bool {
 				continue
 			} else {
 				fields := strings.Fields(lines[i])
-				if _, ok := releases[fields[0]]; !ok {
+				if _, ok := releases[ns]; !ok {
 					releases[ns] = make(map[string]bool)
 				}
 				releases[ns][fields[0][0:strings.LastIndex(fields[0], ".v")]] = true


luisdavim:
This makes sense to me but I think it would be safer if you would check that the `tillerNamespace` set in the state for that release matches the `nameSpace` you're trying to delete from. This would be to avoid the situation where in a multi-tenant environment you would have releases with the same name being managed by different Tillers in different `nameSpaces`.

What do you think?

despairblue:
I'm having a hard time wrapping my head around that. 

In terms of code this would not happen in the `getHelmsmanReleases` function, which I patched to return all helmsman releases (before that it would only return one per namespace because it would recreate the map every time overwriting the old map).
So that would happen in `cleanUntrackedReleases`, right? And isn't that happening [here](https://github.com/Praqma/helmsman/blob/9982b1ef5e79e1aa2d6654bcc56cc33d9457b84b/helm_helpers.go#L411) already?

luisdavim:
I was thinking of doing it in the `getHelmsmanReleases` function, something like `if releases[ns] &&  releases[tillerNamespace] == ns`

despairblue:
Awesome, thank you. 

I think the command `kubectl get configmap -n kube-system -l MANAGED-BY=HELMSMAN` already restricts the releases to the namespace we're checking, `kube-system` in that case.

But maybe I'm not getting it :D

sami-alajrami:
it only checks for untracked releases in namespaces where you use or deploy Tiller as per your DSF (in this case, only `kube-system`). 

So if someone else is using another release with the same name with another Tiller (in a different namespace e.g. `dev`), they will not be impacted.

However, if your DSF defines the namespace `dev` with `installTiller` or `useTiller` set to true, then the cleanup will impact the releases managed by Tiller in the `dev` namespace (other people's work). In this case, you have to use the `--keep-untracked-releases` which will skip the cleanup all together. Using this flag can (in some (possibly) edge cases) leave some releases around which you no longer track in your DSF. But this is better than deleting other people's releases.

I am open for better ways to improve this cleanup process if anyone has suggestions :)

PR #121, opened 2018-11-05
Added environment variable substitution to values and secret files

fsniper:


fsniper:
I also created a pr for global dfs file environment var expansion. 
https://github.com/Praqma/helmsman/pull/122 

I made it as a separate pr because that one is a breaking change

sami-alajrami:
thanks @fsniper could you please update the relevant documentation, at least the [desired state spec](https://github.com/Praqma/helmsman/blob/master/docs/desired_state_specification.md) would need a line on that feature under the `Apps` section.

luisdavim:
Do we need both this and #122 ?

fsniper:
I am focused on the #122 . I am adding some test cases and updating the documentation.

luisdavim:
Yeah, the solution in #122 is more comprehensive and if @sami-alajrami agrees with what's being proposed there we can close this PR.

sami-alajrami:
this can be closed since #122 is now merged.

PR #122, opened 2018-11-05
Breaking: Expanding envs on the whole dfs file.

fsniper:
This patch brings a global dfs file environment variable expansion. For
escaping "$" characters one needs to use "$$".

This change may impact some users with dfs files including "$" strings.

### DIFF
--- a/release.go
+++ b/release.go
@@ -66,7 +66,7 @@ func validateRelease(appLabel string, r *release, names map[string]map[string]bo
 	if r.Version == "" {
 		return false, "version can't be empty."
 	}
-	r.Version = substituteEnv(r.Version)
+	r.Version = r.Version
 
 	_, err := os.Stat(r.ValuesFile)
 	if r.ValuesFile != "" && (!isOfType(r.ValuesFile, ".yaml") || err != nil) {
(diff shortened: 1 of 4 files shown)


utils.go
```diff
@@ -175,12 +181,18 @@ func logVersions() {
 	log.Println("VERBOSE: Helm client version: " + helmVersion)
 }
 
+// add $$ escaping for $ strings
+func expandEnv(s string) string {
+	os.Setenv("HELMSMAN_DOLLAR", "$")
```
luisdavim: Why not do this directly in the `substituteEnv` function bellow?

fsniper: No specific reason. Both can be used.

luisdavim:
@sami-alajrami what's your view on this?
I actually don't think think this is a major braking change, it's only braking if the user had strings with `$`s in the fields that where not being expanded by `substituteEnv` and if they want those `$` signs to be kept unprocessed they need to replace them with `$$` otherwise this will just work as it did before.

release.go
```diff
@@ -66,7 +66,7 @@ func validateRelease(appLabel string, r *release, names map[string]map[string]bo
 	if r.Version == "" {
 		return false, "version can't be empty."
 	}
-	r.Version = substituteEnv(r.Version)
+	r.Version = r.Version
```
sami-alajrami: this line can now be removed.

sami-alajrami:
apologies @fsniper that I didn't have a chance to look at this earlier. It is good. Just one minor redundant line which can be removed and then it is good to merge. We will cut a new release after merging this.

release.go
fsniper: I removed it.

PR #144, opened 2018-12-13
feat: Support local charts on file system

### LINKED ISSUE #51: Support local charts on file system
According to the desired state docs (for the `chart` field) it must contain `repo` and `chart`. 
I would like to be able to deploy a chart from the local file system as well.

Example:

```yaml
apps:
  jenkins:
    name: "mychart
    description: "My Local Chart
    namespace: "staging"
    enabled: true
    chart: "./mychart"
```

SteveRuble:
This adds support for charts that are on the file system rather than in a repo. It works by checking
whether a chart has a repo that's listed in the repos section of the DSF. If it doesn't, the chart
name is treated as a file path instead. If the chart is not present on the file system, or the chart
on the file system is a different version than specified in the DSF, validation fails.

This fixes #51.

### DIFF
--- a/decision_maker.go
+++ b/decision_maker.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"regexp"
 	"strconv"
 	"strings"
 )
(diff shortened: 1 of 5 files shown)


decision_maker.go
```diff
@@ -283,13 +284,19 @@ func logDecision(decision string, priority int) {
...
-	return strings.TrimSpace(strings.Split(releaseChart, "/")[1])
+	m := chartNameExtractor.FindStringSubmatch(releaseChart)
+	if len(m) == 2 {
+		return m[1]
+	}
 
+	return ""
 }
 
+var chartNameExtractor = regexp.MustCompile(`[\\/]([^\\/]+)$`)
```
luisdavim: why a global variable if it's not used anywhere else?

Issue #151, opened 2018-12-22
Support a tillerless setup


luisdavim:
I think it would be cool to support the setup described here: https://rimusz.net/tillerless-helm/

robbert229:
Hey, so perfect timing. I meant to open up an issue on supporting tillerless helm, but that didn't happen because you did it first. Collab?

luisdavim:
Sure, I haven't started working on it, if you have, please goo ahead.

tmc:
Any movement here?

mkubaczyk:
Hi,
I've proposed #259 PR which is an answer for this (and expands the code from #152)

PR #152, opened 2018-12-23
added poc way of supporting tillerless helm via the tiller plugin

robbert229:
I was recently taking a look at helmsman, and one thing preventing me from adopting it is the fact that it doesn't support tillerless helm. I spent some time hacking togethor an example of what tillerless helm might look like in helmsman. The following code does work, but doesn't have any tests. Can we put this through review and figure out how we can integrate this into helmsman?
#151

### DIFF
--- a/command.go
+++ b/command.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"log"
+	"os"
 	"os/exec"
 	"syscall"
 )
@@ -42,6 +43,10 @@ func (c command) exec(debug bool, verbose bool) (int, string) {
 	var stdout, stderr bytes.Buffer
 	cmd.Stdout = &stdout
 	cmd.Stderr = &stderr
+
+	// we need to tell the TILLER to be silent. This will only matter in
+	// tillerless mode.
+	cmd.Env = append(os.Environ(), "HELM_TILLER_SILENT=true")
 
 	if err := cmd.Start(); err != nil {
 		logError("ERROR: cmd.Start: " + err.Error())
--- a/decision_maker.go
+++ b/decision_maker.go
@@ -29,7 +29,7 @@ func makePlan(s *state) *plan {
 func decide(r *release, s *state) {
 	if destroy {
 		if ok, rs := helmReleaseExists(r, ""); ok {
-			deleteRelease(r, rs, s)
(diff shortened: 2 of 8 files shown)


decision_maker.go
```diff
@@ -29,7 +29,7 @@ func makePlan(s *state) *plan {
 func decide(r *release, s *state) {
 	if destroy {
 		if ok, rs := helmReleaseExists(r, ""); ok {
-			deleteRelease(r, rs)
+			deleteRelease(r, rs, s)
```
luisdavim: I think that passing the whole state here is a bit overkill. Maybe it would be better to create a global variable in the init stage or something like that.

robbert229: Just realized that the global settings variable exists. The tillerless variable already exists in global state. I just need to use it.

command.go
```diff
@@ -43,6 +44,10 @@ func (c command) exec(debug bool, verbose bool) (int, string) {
 	cmd.Stdout = &stdout
 	cmd.Stderr = &stderr
 
+	// we need to tell the TILLER to be silent. This will only matter in
+	// tillerless mode.
+	cmd.Env = append(os.Environ(), "HELM_TILLER_SILENT=true")
```
luisdavim: Maybe set it to false if running in debug mode?

luisdavim:
In `main.go` we initialize `helm` shouldn't there be a check to use `--client-only` if running in tillerless mode?

luisdavim:
I think the Dockerfiles should also be updated to include the helm-tiller plugin.

luisdavim:
@robbert229 can you rebase your branch?

luisdavim:
Hey @robbert229 have you given up on this? You were so close...

robbert229:
I have just been soo busy! Will see about finishing this up some time this
week.

luisdavim:
No pressure, just wanted to know. Thanks

eperdeme:
Looking forward to this, I noticed Helmfile just got it merged in yesterday so will be great when this gets merged.
Much easier for us to secure our EKS clusters with this.

mkubaczyk:
Hi,

I've tried to finish it with #259 using code from this branch

sami-alajrami:
since #259  is merged, this PR can be closed now

PR #159, opened 2018-12-31
make settings.kubeContext optional and respect current context

### LINKED ISSUE #153: make settings.kubeContext optional and respect current context in active kubeconfig file
As a kubernetes as a service provider, I want helmsman to respect the `current context` in the active kubeconfig, so that I can use same helmsman DSF on several clusters without an explicit action other than switching between kubeconfigs.

**History**:
We have been using helmsman in an architecture with quite many Kubernetes clusters. In fact we have been providing Kubernetes clusters as a service with some on AKS, EKS, GCP and some on Gardener as Shoots. We have internal tools that pull the kubeconfig files out of the existing systems into a dedicated kubeconfig file and consume it as it is provided from the upstream project. We dont merge them into our local kubeconfig but rather export each into a dedicated kubeconfig file and export KUBECONFIG before calling any kubernetes clients (kubectl, helm, helmsman, etc.). But each service provider have their own internal logic to build a context name in the exported kubeconfig file. helmsman doesn't support using the current context in the exported kubecontext. This forces us to write wrappers around helmsman, that either discovers the current contest, or replicates the context name generation logic of the upstream provider so that we can explicitly pass that to helmsman. What we do is actually creating our own workaround for the lack of the "current context" support in helmsman.

**Bonus**: In most cases we prefer passing explicit `--kubeconfig` argument when calling any kubernetes client rather then implicit KUBECONFIG environment variable. Most upstream tools support this as an explicit flag (e.g. kubectl, helm), we'd like to follow the same practice with helmsman as well.

luisdavim:
This should fix #153 @bergerx could you test it in your use case?

### DIFF
--- a/docs/cmd_reference.md
+++ b/docs/cmd_reference.md
@@ -10,53 +10,56 @@ This is the list of the available CMD options in Helmsman:
 
   `--apply`
         apply the plan directly.
-        
+
   `--apply-labels`
         apply Helmsman labels to Helm state for all defined apps.
-        
+
   `--debug`
         show the execution logs.
-        
+
   `--destroy`
         delete all deployed releases. Purge delete is used if the purge option is set to true for the releases.
-        
+
   `--dry-run`
         apply the dry-run option for helm commands.
-        
+
   `-e value`
         file(s) to load environment variables from (default .env), may be supplied more than once.
-        
+
   `-f value`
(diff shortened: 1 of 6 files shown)


bergerx:
I just tried, worked with both (no `settings.kubeContext` in dsf and with explicit `--kubeconfig`), cool thanks.

Issue #180, opened 2019-01-24
Support cluster-admin tiller roles outside of kube-system


gregsidelinger:
It would be nice if helmsman created tiller service accounts could be bound to cluster-admin when not in the kube-system namespace.  Having to create the namespace, service account and role binding before calling helmsman is kind of annoying.  The current behavior makes sense but I have use cases where I would rather not do a bunch of extra steps before calling helmsman.

flyhard:
Seems like this is related to #125

sami-alajrami:
this is shipped in v1.7.4

PR #182, opened 2019-01-30
Adding support to specify the tiller role.

gregsidelinger:
Addressing Issue https://github.com/Praqma/helmsman/issues/180


flyhard:
Would be even better if you could specify it as a additional role. That would allow for the default role to be applied and you only have to grant access to what is needed in addition.

PR #184, opened 2019-02-03
Add annotation functionality to the namespace

mayppong:
Hi there,

I would like a submit a request for allowing annotations to be added to namespaces specify in the helmsman state file. We use kiam to manage AWS roles for our cluster and it involves annotating roles with allowed roles for the resources to assume. This will let us setup those annotation from within helmsman state file instead of needing to maintain a separate files for the namespaces.

This is my first time writing Go so hopefully this is fine. I pretty much copy from the label function since it should work in a similar fashion from Kubernetes' perspective. I also can't find the test files for kube_helpers, so if one does exists, please point me to it and I can write tests for it as well.

Thank you


luisdavim:
Hi, thanks for your PR, can you please update the documentation and examples?

PR #201, opened 2019-02-21
Add support for externally configured helm repos

turip:
This patch adds support for externally configured helm repositories,
which is a handy tool if you are using Artifcatory as helm repository
server.

In our case, we don't want to push the basic auth credentials to the repositories, but we will make sure that all developers update their helm settings to contain the correct credentials.

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -141,6 +141,15 @@ func toFile(file string, s *state) {
 	}
 }
 
+func stringInSlice(a string, list []string) bool {
+	for _, b := range list {
+		if b == a {
+			return true
+		}
+	}
+	return false
+}
+
 func resolvePaths(relativeToFile string, s *state) {
 	dir := filepath.Dir(relativeToFile)
 
@@ -161,6 +170,7 @@ func resolvePaths(relativeToFile string, s *state) {
 		if v.Chart != "" {
 			var repoOrDir = filepath.Dir(v.Chart)
 			_, isRepo := s.HelmRepos[repoOrDir]
+			isRepo = isRepo || stringInSlice(repoOrDir, s.PreconfiguredHelmRepos)
 			if !isRepo {
 				// if there is no repo for the chart, we assume it's intended to be a local path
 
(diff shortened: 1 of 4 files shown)


turip:
If you are not agreeing with this approach there are other ways around this, (like providing a global config for secrets), however, your input would be appreciated so that I can create a patch that is acceptable into helmsman.

utils.go
```diff
@@ -141,6 +141,15 @@ func toFile(file string, s *state) {
 	}
 }
 
+func stringInSlice(a string, list []string) bool {
+	for _, b := range list {
```
luisdavim: No super important here because these slices will always be small but it would be better to use something like `sort.SearchStrings` instead of a linear search.

turip: Yep that's possible but for me given that here's no performance implication, this one is even harder to read:

```go

sort.Sort(StringSlice(s.PreconfiguredHelmRepos))

possible_index = sort.SearchStrings(s.PreconfiguredHelmRepos, repoOrDir)
isRepo = isRepo || (possible_index < len(s.PreconfiguredHelmRepos) && s.PreconfiguredHelmRepos[possible_index] == repoOrDir)
```
The other thing that can happen is to use a map[string]bool type, and instead of the PreconfiguredHelmRepos have something like:

```yaml
manageHelmRepos:
  my-repo: false
```

This would mean that:
1) by default all helm repos are managed that does not have this key
2) if the user sets it to true then those repos are still managed
3) if it's set to false then it's an opt out

But I would prefer to go with this changeset for now, unless you insist on me implementing this approach.

turip:
Could someone please merge this PR, it would be extremely helpful for us to have this functionality.

luisdavim:
Only @sami-alajrami can merge PRs and I'm sure he'll get around to it soon no need for pressure.

turip:
Sure, I am sorry, for that.  I was just interested in how this is done with this project.

sami-alajrami:
thanks @turip for this and apologies for the delay!

PR #204, opened 2019-02-25
Make settings optional in desired state file.

### LINKED ISSUE #203: Make settings optional
I'm trying to let helmsman determine the kube context from current-context. I cannot run this without atleast one setting specified in the dsf. This works to me:
```
settings:
  password: "asdf"
```
while this doesn't work:
```
settings:
 # password: "asdf"
```

erihanse:
Now that helmsman can use kube config current-context there's no reason anymore for settings being mandatory. This pull request aims to change this behaviour. 

Fixes #203

Signed-off-by: Erik Aaron Hansen <erihanse@outlook.com>

### DIFF
--- a/docs/desired_state_specification.md
+++ b/docs/desired_state_specification.md
@@ -73,12 +73,12 @@ certificates:
 
 ## Settings
 
-Optional : No.
+Optional : Yes.
 
 Synopsis: provides settings for connecting to your k8s cluster and configuring Helm's Tiller in the cluster.
 
 Options:
-- kubeContext : this is always required and defines what context to use in kubectl. Helmsman will try connect to this context first, if it does not exist, it will try to create it (i.e. connect to a k8s cluster) using the options below.
+- kubeContext : optionally define what context to use in kubectl. Helmsman will try connect to this context first, if it does not exist, it will try to create it (i.e. connect to a k8s cluster) using the options below.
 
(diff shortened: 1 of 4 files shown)


erihanse:
I'm sorry, not sure why the build failed. Tests made it through on my laptop with `make test`.

luisdavim:
You need to update the test that says that settings can't be empty:
```
--- FAIL: Test_state_validate/test_case_2_--_settings/nil_value (0.06s)
    	state_test.go:453: state.validate() = false, want true
```

erihanse:
Hmmm. I did that. I just cannot find out what's wrong. When running `make test` locally case 2 for state validation returns true. 
```
--- PASS: Test_state_validate (0.16s)
    --- PASS: Test_state_validate/test_case_1 (0.00s)
    --- PASS: Test_state_validate/test_case_2_--_settings/nil_value (0.08s)
```
however, on the build server it returns false.

erihanse:
Ah, it's probably because I have a kube context compared to the test environment on the build server.

sami-alajrami:
thanks @erihanse .. I have removed the empty settings validation error in #206 before seeing your PR. 
The test passes on your machine because you have a kubecontext configured ( I am guessing) so the validation returns true. 
In CI, there is no kubecontext configured and therefore the validation returns false (correct behaviour). Since it's somewhat troublesome test, I have removed in #206

erihanse:
Okay, you can remove this pull request then :)

PR #205, opened 2019-02-26
Minor document formatting fixes

spacez320:
Just making the desired spec docs look a little more nicer.


sami-alajrami:
thanks @spacez320 :)

Issue #207, opened 2019-02-28
AWS IAM machine credentials are not used


ironhalo:
I'm running helmsman in CD as a container on an ec2 instance while storing helm TLS certs in S3.  Helmsman requires the AWS credentials to be exposed as environmental variables and does not seem to respect the credentials presented by the IAM role attached to the instance.  My workaround have been to extract the credentials from the metadata endpoint, is it possible to use the already available credentials natively?

ironhalo:
Looks like the source of the error is from this function?

https://github.com/Praqma/helmsman/blob/master/aws/aws.go#L38

sami-alajrami:
should be fixed in v1.8.1 @ironhalo please reopen if the issue persists.

PR #208, opened 2019-03-01
Remove AWS environment variable check.

jkrukoff-cb:
The AWS SDK is capable of sourcing environment variables from a variety
of sources. The removed check prevents it from doing so for anything but
environment variables.

In particular, we'd like to get EC2 machine roles to work.

This is a possible fix for https://github.com/Praqma/helmsman/issues/207

### DIFF
--- a/aws/aws.go
+++ b/aws/aws.go
@@ -36,7 +36,7 @@ func ReadFile(bucketName string, filename string, outFile string, noColors bool)
 	style = aurora.NewAurora(!noColors)
 	// Checking env vars are set to configure AWS
 	if !checkCredentialsEnvVar() {
-		log.Fatal(style.Bold(style.Red("ERROR: Failed to find the AWS env vars needed to configure AWS. Please make sure they are set in the environment.")))
+		log.Println("WARN: Failed to find the AWS env vars needed to configure AWS. Please make sure they are set in the environment.")
 	}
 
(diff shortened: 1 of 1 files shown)


aws/aws.go
```diff
@@ -14,30 +14,9 @@ import (
...
-	}
-	return true
-}
-
 // ReadFile reads a file from S3 bucket and saves it in a desired location.
 func ReadFile(bucketName string, filename string, outFile string, noColors bool) {
 	style = aurora.NewAurora(!noColors)
-	// Checking env vars are set to configure AWS
-	if !checkCredentialsEnvVar() {
-		log.Fatal(style.Bold(style.Red("ERROR: Failed to find the AWS env vars needed to configure AWS. Please make sure they are set in the environment.")))
```
sami-alajrami: thanks @jkrukoff-cb for the PR .. since the validation already sets AWS_REGION env var to support for users who have AWS_DEFAULT_REGION set instead, removing it can introduce a minor breaking change. 
I think it would be better to make the validation optional, i.e. replace the log.Fatal with log.Println saying that no environment variables are being used. Then the sdk can use any other alternatives that are available in the machine.

jkrukoff-cb:
Thanks for the guidance, have updated PR to convert this into a warning as requested.

sami-alajrami:
Great! thanks again @jkrukoff-cb :)

PR #218, opened 2019-03-16
Dehardcoding LimitRange type to be able to set limits for Pod

hatemosphere:
There is an outstanding issue in Kubernetes docs about missing Pod type in LimitRange doc - https://github.com/kubernetes/website/issues/9585, so i guess that's why it wasn't added. Although this type exists and we are successfully using it for some time. The proof can be found here for example - https://docs.openshift.com/container-platform/3.11/dev_guide/compute_resources.html#dev-limit-ranges

### DIFF
--- a/docs/desired_state_specification.md
+++ b/docs/desired_state_specification.md
@@ -75,7 +75,7 @@ certificates:
 
 ## Settings
 
-Optional : Yes. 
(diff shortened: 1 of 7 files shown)


hatemosphere:
let me know if you think this should be backward compatible

PR #240, opened 2019-04-26
substitute env vars in values files. fixes #226

### LINKED ISSUE #226: Env substitution in values files
Hello,

Is it possible to use the ${} environment variables inside the values Files?

I've checked the issues list and it looks like https://github.com/Praqma/helmsman/pull/121 was not merged, which looked to run the env parser against all the values files.

I'd prefer to keep all the config in the values Files rather than in the DSF

So rather than having to do some stuff via set and others via valuesFile;

`````
  autoscaler:
    namespace: kube-system
    enabled: true
    wait: true
    chart: stable/cluster-autoscaler
    version: 0.12.1
    priority: -100
    set:
      autoDiscovery.clusterName: ${kube_cluster_name}
    valuesFiles:
      - "helmsman/autoscaler.yaml"
`````

I think it's cleaner to do

helmsman/autoscaler.yaml
````yaml
autoDiscovery:
  clusterName: ${kube_cluster_name}
sslCertPath: /etc/kubernetes/pki/ca.crt
rbac:
  create: true
cloudProvider: aws
image:
  tag: v1.12.3

````

### PR DESCRIPTION
**Pull request by sami-alajrami**
this replicates behaviour in #121 to fix #226

### DIFF
--- a/helm_helpers.go
+++ b/helm_helpers.go
@@ -381,7 +381,6 @@ func deployTiller(namespace string, serviceAccount string, defaultServiceAccount
 	tls := ""
 	ns := s.Namespaces[namespace]
 	if tillerTLSEnabled(ns) {
-		shouldCleanup = true // needed to activate cleaning up the certs upon exit
 		tillerCert := namespace + "-tiller.cert"
 		tillerKey := namespace + "-tiller.key"
 		caCert := namespace + "-ca.cert"
--- a/main.go
+++ b/main.go
@@ -30,7 +30,6 @@ var noColors bool
 var noFancy bool
 var noNs bool
 var nsOverride string
-var shouldCleanup bool
 var skipValidation bool
 var applyLabels bool
 var keepUntrackedReleases bool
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by eperdeme** [comment]
Oh thanks! I've been trying to work around this but my yaml files are getting so messy now with half the values in helmsman.yaml if it contains ENVS and half in app/values.yaml for staticit's got very messy.

**Comment by eperdeme** [comment]
Any chance you can merge this in so we can switch back to official binaries ? :)

**Comment by sami-alajrami** [comment]
@eperdeme I didn't get a chance to test it yet, can you verify that it is working as intended ? :)

**Comment by eperdeme** [comment]
Had a go but it seems to not substituteEnv anything in the "valuesFiles:" .

I'll try and take a look but my Go is limited.

**Comment by eperdeme** [comment]
If I had to guess without understanding GO, I'd say the current pull if trying to do substituteEnv against a list of paths and not the file contents.
I'd guess we need a  ioutil.ReadFile(file) loop around the file paths.

**Comment by sami-alajrami** [comment]
to avoid messing up the original files (containing env vars), I changed the behaviour to make a temp copy of the file and use it then delete it at the end of execution.

**Comment by xaka** [comment]
Turned out there is a problem when multiple files with the same base name are passed in:
```
      valuesFiles:
        - ../helmsman.values.yaml
        - helmsman.values.yaml
```

It'll lead to the following helm command:
```
helm -f tmp/helmsman.values.yaml -f tmp/helmsman.values.yaml
```

It might be better to generate uniq names for each output file.

PR #242, opened 2019-05-04
Improve temporary dir handling.

brndnmtthws:
- renamed temp dir from `helmsman-temp-dir` to `.helmsman-tmp`
 - defer cleanup of temporary files at the start of main()
 - use ioutil TempDir() instead of custom code

There's still one outstanding issue which is that the temp dir won't be
cleaned up properly if the program exits early while if an error happens
inside init().

I had an issue with deployments flapping, and it turned out the cause was a conflict in the temporary directory. This change puts each file into a separate path.

### DIFF
--- a/init.go
+++ b/init.go
@@ -133,6 +133,10 @@ func init() {
 		}
 	}
 
+	// wipe & create a temporary directory
+	os.RemoveAll(tempFilesDir)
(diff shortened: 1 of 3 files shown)


sami-alajrami:
@brndnmtthws thanks for your contribution :)

PR #248, opened 2019-05-15
Do not delete untracked releases if the target flag is set.

fmotrifork:
The new  `-target` flag does not ignore untracked releases, even when specifically stating which release we are targeting.

Releases that are not specified by a  `-target` flag is now also ignored.

I have added tests for the `-target` flag behaviour defined in the `decision_maker.go` file, but I am unfortunately not able to write a test for `-target` flag for untracked releases, in the way the code is structured today.

I was thinking that maybe the `func cleanUntrackedReleases()` in `helm_helper` should actually be a part of the `decide(r,s)` func so that all decisions are made when building the plan?

### DIFF
--- a/decision_maker_test.go
+++ b/decision_maker_test.go
@@ -76,3 +76,118 @@ func Test_getValuesFiles(t *testing.T) {
 		})
 	}
 }
+
+func Test_decide(t *testing.T) {
+	type args struct {
+		r *release
+		s *state
+	}
+	tests := []struct {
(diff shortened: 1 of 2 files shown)


fmotrifork:
@sami-alajrami Do you have a minute to look at this PR? :slightly_smiling_face:

sami-alajrami:
@fmotrifork apologies it took so long to get around it. Thank you for the contribution!

Issue #250, opened 2019-05-28
"helm diff upgrade ..." is missing helmFlags


xaka:
`upgradeRelease` function uses `helmFlags` while `diffRelease` doesn't (https://github.com/Praqma/helmsman/blob/master/decision_maker.go#L245)

PR #252, opened 2019-05-29
add helm flags to "helm diff upgrade ..." command

xaka:
fixes https://github.com/Praqma/helmsman/issues/250


xaka:
If possible, I'd love it to be back-ported to 1.8.x and 1.8.2 released.

PR #255, opened 2019-06-03
Consider -target flag while running untracked releases purging

### LINKED ISSUE #254: Helmsman is removing untracked releases even with -target flag
Even though -target flag is set to chart X, helmsman removes untracked Y release.
I would suppose targeting would scope it to specific chart only no matter what step of helmsman invoking it is.

And since I'm aware of -keep-untracked-releases flag, I still think this should not be required when target is set, because setting proper scope should have a precedence over such flag.

mkubaczyk:
Resolves #254

### DIFF
--- a/helm_helpers.go
+++ b/helm_helpers.go
@@ -472,10 +472,14 @@ func cleanUntrackedReleases() {
 				}
 			}
 			if !tracked {
-				if _, ok := toDelete[ns]; !ok {
-					toDelete[ns] = make(map[string]bool)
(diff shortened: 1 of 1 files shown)


fmotrifork:
This solves the same problem as https://github.com/Praqma/helmsman/pull/248
I havn't looked at which implementation is the cleanest.

sami-alajrami:
thanks @mkubaczyk  .. both this PR and #248 do fix the issue .. however, the latter PR adds some tests and therefore I would go for merging #248  and closing this one. Thank you both for the fixes 🎉

PR #258, opened 2019-06-08
Add TillerRoleTemplateFile to allow create Tiller with custom role

mkubaczyk:
My motivation was the issue I faced once I tried to split applications into different namespaces. Some of them perfectly fit into the default data/role.yaml, but some required namespace-specific permissions and I didn't want to create such roles by hand or outside of a helmsman. Hence I'm coming up with a PR with implementation of TillerRoleConfigFile field, which allows to point a yaml template file (similar to data/role.yaml) where custom namespaced helmsman role can be defined.


mkubaczyk:
I've renamed tillerRoleConfigFile to tillerRoleTemplateFile

helm_helpers.go
```diff
@@ -342,28 +342,46 @@ func addHelmRepos(repos map[string]string) (bool, string) {
...
 					return false, rbacErr
 				}
 			} else {
 				return false, "ERROR: while validating/creating service account [ " + serviceAccount + " ] in namespace [" + namespace + "]: " + err
 			}
 		}
 		sa = "--service-account " + serviceAccount
+	} else if serviceAccount == "" && roleTemplateFile != "" {
+		roleName := "helmsman-tiller"
+		defaultServiceAccountName := "helmsman"
```
sami-alajrami: Thank you @mkubaczyk for this PR .. I have wanted to add this for a long time but didn't get around to it :)
I have a small comment on the `defaultServiceAccountName` here .. I am wondering why you reset here even though it might hold a value which is passed to the deployTiller function. 

I would suggest to check if defaultServiceAccountName has a value and use it, otherwise, set its value to helmsman. This way, you can also get rid of the third else if below in line 379.

mkubaczyk: Hi, thanks for an idea. I did what you've asked for. Code in fact became a lot easier in this function. :-)

mkubaczyk:
I've rebased my branch onto master to resolve conflicts in helm_helpers.go

PR #259, opened 2019-06-08
Add support for Tillerless deployment made with helm-tiller plugin

### LINKED ISSUE #151: Support a tillerless setup

mkubaczyk:
Resolves #151, adds support for helm-tiller plugin, which is in fact a Tillerless deployment in V2 helm.
It is an answer for #152 PR which seems to be abandoned. I've taken the code from it and moved it the reasonable end. All kudos to the original author of #152.


sami-alajrami:
Awesome! thanks @robbert229 and @mkubaczyk  for your efforts on this!!

PR #260, opened 2019-06-09
Fix docs

nvtkaszpir:
Fix broken links in the documnetation. Some documents were moved
without updating references in other documents.

Some formatting updates.


sami-alajrami:
thanks @nvtkaszpir for these fixes :)

PR #265, opened 2019-06-27
Remove getHelmFlags from diff, add to install/upgrade/delete release func

### LINKED ISSUE #264: Problems with helmFlags in v1.10.0
As of v1.10.0 user-supplied `helmFlags` are no longer passed to `installRelease()`, `upgradeRelease()`, or `reInstallRelease()`. This comes from changes in #259, perhaps as a result of adapting an older PR.

Additionally, #252 effectively added a number of extra flags including `--timeout` to `helm diff upgrade`. This is not supported by helm-diff and results in the following error if `timeout` is set in the DSF:
```
Command returned with exit code: . And error message: Error: unknown flag: --timeout
```

Since very few flags are common to all helm commands, I would suggest we need to split these into separate arrays (`helmFlags` and `helmDiffFlags`, for example).

mkubaczyk:
Should fix #264

### DIFF
--- a/decision_maker.go
+++ b/decision_maker.go
@@ -137,7 +137,7 @@ func installRelease(r *release) {
 
 	cmd := command{
 		Cmd:         "bash",
-		Args:        []string{"-c", helmCommandFromConfig(r) + " install " + r.Chart + " -n " + r.Name + " --namespace " + r.Namespace + getValuesFiles(r) + " --version " + r.Version + getSetValues(r) + getSetStringValues(r) + getWait(r) + getDesiredTillerNamespaceFlag(r) + getTLSFlags(r) + getTimeout(r) + getNoHooks(r) + getDryRunFlags()},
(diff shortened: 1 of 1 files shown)


mkubaczyk:
> `getHelmFlags()` includes all the flags from `getNoHooks()`, `getTimeout()` and `getDryRunFlags()`. Those should be removed either from that function or from these individual Args strings, to avoid duplicates.

Not really. As far as I can tell, those three have their own fields to use, while helmFlags accept anything in form of list. I believe it is users' duty to not use things in helmFlags that have their own dedicated fields.

lachlancooper:
What I mean is the `getHelmFlags()` function literally returns those other three sets of flags as well as appending any user-provided `helmFlags`. You can see this here: https://github.com/mkubaczyk/helmsman/blob/4aad7c76336bcd85ce31cb6f823ca6d1534094bd/decision_maker.go#L521

mkubaczyk:
Oh, I’ve not seen that and I’ve been looking on this function couple of times. Ok, I’ll sit again to this, maybe with some additional tests written this time

PR #281, opened 2019-08-13
Add ability to disable environment substitution in values files

### LINKED ISSUE #280: Ability to disable env substitution only in value files
Env variable substitution in desired state files is a very useful feature and I use it for dynamically generated test environments, but now when the substitution has been extended to the values files it causes problems with accidental value substitution (e.g. a lot of randomly generated passwords and command definitions contain `$` sign). I would be cool to disable substitution in value files so the state file could remain dynamic and values files static.

I can prepare PR if this functionality is acceptable.

medzin:
Add ability to disable environment substitution in values files.

Resolves #280

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -55,7 +55,7 @@ func fromTOML(file string, s *state) (bool, string) {
 	}
 	addDefaultHelmRepos(s)
 	resolvePaths(file, s)
-	if !noEnvSubst {
+	if !noEnvSubst && !noEnvValuesSubst {
 		substituteEnvInValuesFiles(s)
 	}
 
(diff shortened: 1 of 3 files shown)


mkubaczyk:
lgtm +1

utils.go
```diff
@@ -55,7 +55,7 @@ func fromTOML(file string, s *state) (bool, string) {
 	}
 	addDefaultHelmRepos(s)
 	resolvePaths(file, s)
-	if !noEnvSubst {
+	if !noEnvSubst || !noEnvValuesSubst {
```
sami-alajrami: I think this should be `&&` .. the current condition means that if `noEnvSubst = false` and `noEnvValuesSubst = true` , environment variables will still be substituted in values files.

utils.go
```diff
@@ -107,7 +107,7 @@ func fromYAML(file string, s *state) (bool, string) {
 	addDefaultHelmRepos(s)
 	resolvePaths(file, s)
 
-	if !noEnvSubst {
+	if !noEnvSubst || !noEnvValuesSubst {
```
sami-alajrami: same as above

utils.go
medzin: Yep, you are right - thanks.

medzin:
@sami-alajrami fixed!

PR #287, opened 2019-08-19
Added support of running Helmsman on Windows

PeterBrun:
Any execution of a command using bash with the command parameter has been replaced with just the expected command executable (helm and kubectl). And instead of building the arguments as a single string, they are now built using string arrays. This solves the "quoting" problem, that is there when using a single string with all arguments. For example, before this change, a path had to be quoted on Windows to work. By just building string arrays with all the arguments, this problem is solved by Go' os/exec package.

It also solves the problem of using a password protected Helm repo, where the username contains a '$' (for example Harbor creates robot accounts with this). This could be escaped using '$$', but then it still was being interpreted by bash as an environment variable. This could have been solved by adding quotes around in the old solution, but it is just simpler to just use the "native" string array arguments approach as described above. And still if the environment variable was intended as input, then it is already applied by the "substituteEnv" functionality.

### DIFF
--- a/decision_maker.go
+++ b/decision_maker.go
@@ -3,6 +3,7 @@ package main
 import (
 	"fmt"
 	"regexp"
+	"runtime"
 	"strconv"
 	"strings"
 )
@@ -107,29 +108,33 @@ func decide(r *release, s *state) {
 // operate without a tiller it will return `helm tiller run NAMESPACE -- helm`
 // where NAMESPACE is the namespace that the release is configured to use.
 // If not configured to run without a tiller will just return `helm`.
-func helmCommand(namespace string) string {
+func helmCommand(namespace string) []string {
 	if settings.Tillerless {
+		if runtime.GOOS == "windows" {
+			logError("ERROR: Tillerless Helm plugin is not supported on Windows")
+		}
+		return []string{"tiller", "run", namespace, "--", "helm"}
 		if namespace != "" {
-			return "helm tiller run " + namespace + " -- helm"
+			return []string{"tiller", "run", namespace, "--", "helm"}
 		} else {
-			return "helm tiller run helm"
+			return []string{"tiller", "run", "helm"}
 		}
 	}
 
-	return "helm"
+	return nil
 }
 
 // helmCommandFromConfig calls helmCommand returning the correct way to invoke
 // helm.
-func helmCommandFromConfig(r *release) string {
+func helmCommandFromConfig(r *release) []string {
 	return helmCommand(getDesiredTillerNamespace(r))
 }
 
 // testRelease creates a Helm command to test a particular release.
 func testRelease(r *release) {
 	cmd := command{
-		Cmd:         "bash",
-		Args:        []string{"-c", helmCommandFromConfig(r) + " test " + r.Name + getDesiredTillerNamespaceFlag(r) + getTLSFlags(r)},
+		Cmd:         "helm",
+		Args:        concat(helmCommandFromConfig(r), []string{"test", r.Name}, getDesiredTillerNamespaceFlag(r), getTLSFlags(r)),
 		Description: generateCmdDescription(r, "running tests for"),
 	}
 	outcome.addCommand(cmd, r.Priority, r)
@@ -139,8 +144,8 @@ func testRelease(r *release) {
 // installRelease creates a Helm command to install a particular release in a particular namespace using a particular Tiller.
 func installRelease(r *release) {
 	cmd := command{
-		Cmd:         "bash",
-		Args:        []string{"-c", helmCommandFromConfig(r) + " install " + r.Chart + " -n " + r.Name + " --namespace " + r.Namespace + getValuesFiles(r) + " --version " + strconv.Quote(r.Version) + getSetValues(r) + getSetStringValues(r) + getWait(r) + getDesiredTillerNamespaceFlag(r) + getTLSFlags(r) + getHelmFlags(r)},
+		Cmd:         "helm",
+		Args:        concat(helmCommandFromConfig(r), []string{"install", r.Chart, "-n", r.Name, "--namespace", r.Namespace}, getValuesFiles(r), []string{"--version", r.Version}, getSetValues(r), getSetStringValues(r), getWait(r), getDesiredTillerNamespaceFlag(r), getTLSFlags(r), getHelmFlags(r)),
 		Description: generateCmdDescription(r, "installing"),
 	}
 	outcome.addCommand(cmd, r.Priority, r)
@@ -159,8 +164,8 @@ func rollbackRelease(r *release, rs releaseState) {
 	if r.Namespace == rs.Namespace {
 
 		cmd := command{
-			Cmd:         "bash",
-			Args:        []string{"-c", helmCommandFromConfig(r) + " rollback " + r.Name + " " + getReleaseRevision(rs) + getWait(r) + getDesiredTillerNamespaceFlag(r) + getTLSFlags(r) + getTimeout(r) + getNoHooks(r) + getDryRunFlags()},
+			Cmd:         "helm",
+			Args:        concat(helmCommandFromConfig(r), []string{"rollback", r.Name, getReleaseRevision(rs)}, getWait(r), getDesiredTillerNamespaceFlag(r), getTLSFlags(r), getTimeout(r), getNoHooks(r), getDryRunFlags()),
 			Description: generateCmdDescription(r, "rolling back"),
 		}
 		outcome.addCommand(cmd, r.Priority, r)
@@ -193,8 +198,8 @@ func deleteRelease(r *release, rs releaseState) {
 	}
 
 	cmd := command{
-		Cmd:         "bash",
-		Args:        []string{"-c", helmCommandFromConfig(r) + " delete " + p + " " + r.Name + getCurrentTillerNamespaceFlag(rs) + getTLSFlags(r) + getDryRunFlags()},
+		Cmd:         "helm",
+		Args:        concat(helmCommandFromConfig(r), []string{"delete", p, r.Name}, getCurrentTillerNamespaceFlag(rs), getTLSFlags(r), getDryRunFlags()),
 		Description: generateCmdDescription(r, "deleting"),
 	}
 	outcome.addCommand(cmd, priority, r)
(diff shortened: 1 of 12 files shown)


decision_maker.go
```diff
@@ -212,12 +216,14 @@ func inspectUpgradeScenario(r *release, rs releaseState) {
 
 	if r.Namespace == rs.Namespace {
 
-		version, msg := getChartVersion(r)
-		if msg != "" {
-			logError(msg)
-			return
+		if !isLocalChart(r.Chart) {
```
nvanheuverzwijn: This part of the patch should go into another PR and be merged as soon as possible.

psbtrifork: I have removed this patch from this PR. Awaiting PR #291 instead.

psbtrifork: I have removed this patch from this PR. Awaiting PR #291 instead.

PeterBrun: I have removed this patch from this PR. Awaiting PR #291 instead.

nvanheuverzwijn:
@luisdavim any reason why commands used to be call with `bash -c` instead of using `helm` directly ? This PR changes how helm is called, not sure if it has any impact or not.

decision_maker.go
```diff
@@ -393,36 +397,37 @@ func getValuesFiles(r *release) string {
...
 }
 
 // getSetValues returns --set params to be used with helm install/upgrade commands
-func getSetValues(r *release) string {
-	result := ""
+func getSetValues(r *release) []string {
+	result := []string{}
 	for k, v := range r.Set {
-		result = result + " --set " + k + "=\"" + strings.Replace(v, ",", "\\,", -1) + "\""
+		result = append(result, "--set", k+"="+strings.Replace(v, ",", "\\,", -1)+"")
```
nvanheuverzwijn: This here seems to be different. The resulting string would be `k=something` instead of k="something". It would break if `v` have any space. Unless `v` is never supposed to have space in it.

decision_maker.go
```diff
@@ -393,36 +397,37 @@ func getValuesFiles(r *release) string {
...
 }
 
 // getSetStringValues returns --set-string params to be used with helm install/upgrade commands
-func getSetStringValues(r *release) string {
-	result := ""
+func getSetStringValues(r *release) []string {
+	result := []string{}
 	for k, v := range r.SetString {
-		result = result + " --set-string " + k + "=\"" + strings.Replace(v, ",", "\\,", -1) + "\""
+		result = append(result, "--set-string", k+"="+strings.Replace(v, ",", "\\,", -1)+"")
```
nvanheuverzwijn: Same here.

helm_helpers.go
```diff
@@ -254,12 +255,11 @@ func getReleaseChartVersion(rs releaseState) string {
...
 // getNSTLSFlags returns TLS flags for a given namespace if it's deployed with TLS
-func getNSTLSFlags(namespace string) string {
-	tls := ""
+func getNSTLSFlags(namespace string) []string {
+	tls := []string{}
 	ns := s.Namespaces[namespace]
 	if tillerTLSEnabled(ns) {
-
-		tls = " --tls --tls-ca-cert " + namespace + "-ca.cert --tls-cert " + namespace + "-client.cert --tls-key " + namespace + "-client.key "
+		tls = append(tls, "--tls", "--tls-ca-cert", namespace+"-ca.cert", "--tls-cert", namespace+"-client.cert", "--tls-key", namespace+"-client.key")
```
nvanheuverzwijn: For consistency, I would add space between `+`

helm_helpers.go
```diff
@@ -325,8 +325,8 @@ func getChartVersion(r *release) (string, string) {
 		return r.Version, ""
 	}
 	cmd := command{
-		Cmd:         "bash",
-		Args:        []string{"-c", "helm search " + r.Chart + " --version " + strconv.Quote(r.Version)},
+		Cmd:         "helm",
+		Args:        []string{"search", r.Chart, "--version", r.Version},
```
nvanheuverzwijn: Here, `strconv.Quote` is removed. I am not too sure what the outcome of doing this is.

helm_helpers.go
```diff
@@ -395,13 +395,13 @@ func addHelmRepos(repos map[string]string) (bool, string) {
...
-			basicAuth = " --username " + u.User.Username() + " --password " + p
+			basicAuthArgs = append(basicAuthArgs, "--username", u.User.Username(), "--password", p)
 
 		}
 
 		cmd := command{
-			Cmd:         "bash",
-			Args:        []string{"-c", "helm repo add " + basicAuth + " " + repoName + " " + strconv.Quote(repoLink)},
+			Cmd:         "helm",
+			Args:        concat([]string{"repo", "add", repoName, repoLink}, basicAuthArgs),
```
nvanheuverzwijn: Same here, `strconv.Quote` removed.

helm_helpers.go
```diff
@@ -506,8 +506,8 @@ func deployTiller(namespace string, serviceAccount string, defaultServiceAccount
 // initHelmClientOnly initializes the helm client only (without deploying Tiller)
 func initHelmClientOnly() (bool, string) {
 	cmd := command{
-		Cmd:         "bash",
-		Args:        []string{"-c", "helm init --client-only "},
+		Cmd:         "helm",
+		Args:        []string{"init", "--client-only"},
```
nvanheuverzwijn: There used to be a space after `--client-only` (`"--client-only"` instead of `"--client-only "`). Not sure what the impact is, might be a typo that this PR fixes now.

helm_helpers_test.go
```diff
@@ -8,11 +8,11 @@ import (
...
 	t.Log("setup test case")
-	os.MkdirAll("/tmp/helmsman-tests/myapp", os.ModePerm)
-	os.MkdirAll("/tmp/helmsman-tests/dir-with space/myapp", os.ModePerm)
+	os.MkdirAll(os.TempDir()+"/helmsman-tests/myapp", os.ModePerm)
+	os.MkdirAll(os.TempDir()+"/helmsman-tests/dir-with space/myapp", os.ModePerm)
 	cmd := command{
-		Cmd:         "bash",
-		Args:        []string{"-c", "helm create  '/tmp/helmsman-tests/dir-with space/myapp'"},
+		Cmd:         "helm",
+		Args:        []string{"create", os.TempDir() + "/helmsman-tests/dir-with space/myapp"},
```
nvanheuverzwijn: Single quote `'` are removed here.

helm_helpers_test.go
```diff
@@ -138,7 +138,7 @@ func Test_validateReleaseCharts(t *testing.T) {
 						Namespace:       "",
 						Enabled:         true,
 						Chart:           "stable/prometheus",
-						Version:         "",
+						Version:         "*",
```
nvanheuverzwijn: Why change `Version` from `""` to `"*"` ? Did this PR breaks this test ?

helm_helpers_test.go
```diff
@@ -233,7 +233,7 @@ func Test_validateReleaseCharts(t *testing.T) {
 				targetMap[target] = true
 			}
 			if got, msg := validateReleaseCharts(tt.args.apps); got != tt.want {
-				t.Errorf("getReleaseChartName() = %v, want %v , msg: %v", got, tt.want, msg)
+				t.Errorf("validateReleaseCharts() = %v, want %v , msg: %v", got, tt.want, msg)
```
nvanheuverzwijn: Does this change have anything to do with windows ? Is it because the test did not report the correct behavior ?

helm_helpers_test.go
```diff
@@ -334,7 +334,7 @@ func Test_getReleaseChartVersion(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			t.Log(tt.want)
 			if got := getReleaseChartVersion(tt.args.r); got != tt.want {
-				t.Errorf("getReleaseChartName() = %v, want %v", got, tt.want)
+				t.Errorf("getReleaseChartVersion() = %v, want %v", got, tt.want)
```
nvanheuverzwijn: Does this change have anything to do with windows ? Is it because the test did not report the correct behavior ?

kube_helpers.go
```diff
@@ -230,23 +230,23 @@ func createContext() (bool, string) {
...
-		setCredentialsCmd = "kubectl config set-credentials " + s.Settings.Username + " --token=" + token
+		setCredentialsCmdArgs = append(setCredentialsCmdArgs, "config", "set-credentials", s.Settings.Username, "--token="+token)
 	} else {
-		setCredentialsCmd = "kubectl config set-credentials " + s.Settings.Username + " --username=" + s.Settings.Username +
-			" --password=" + s.Settings.Password + " --client-key=" + caKey
+		setCredentialsCmdArgs = append(setCredentialsCmdArgs, "config", "set-credentials", s.Settings.Username, "--username="+s.Settings.Username,
+			"--password="+s.Settings.Password, "--client-key="+caKey)
 		if caClient != "" {
-			setCredentialsCmd = setCredentialsCmd + " --client-certificate=" + caClient
+			setCredentialsCmdArgs = append(setCredentialsCmdArgs, "--client-certificate="+caClient)
```
nvanheuverzwijn: I would add space between `+` for consistency.

decision_maker.go
PeterBrun: No, it does not break if you enter a space.

decision_maker.go
PeterBrun: No, it does not break if you enter a space.

decision_maker.go
nvanheuverzwijn: Gotcha, just making sure.

helm_helpers.go
PeterBrun: That is how gofmt formats that kind of line. There is over 100 examples of this kind in the existing codebase.

helm_helpers.go
PeterBrun: Please read my description:

> This solves the "quoting" problem, that is there when using a single string with all arguments.

helm_helpers.go
PeterBrun: Please read my description:

> This solves the "quoting" problem, that is there when using a single string with all arguments.

helm_helpers_test.go
PeterBrun: That is because the 'qoute' hack is no longer needed.

helm_helpers_test.go
PeterBrun: I will update the "helm search" code to check for the empty string.

helm_helpers_test.go
PeterBrun: I think the old code is wrong. Would you agree? Especially when debugging a unit test, it says something that is not tested here. Here "validateReleaseCharts" and now it says "validateReleaseCharts" failed if it fails.

helm_helpers_test.go
PeterBrun: I think the old code is wrong. Would you agree? Especially when debugging a unit test, it says something that is not tested here. Here "getReleaseChartVersion" and now it says "getReleaseChartVersion" failed if it fails.

kube_helpers.go
PeterBrun: That is how gofmt formats that kind of line. There is over 120 examples of this kind in the existing code base.

decision_maker.go
nvanheuverzwijn: Refactors are hard and sometimes, small details will destroy the effort, just making sure that this particular ommision of the double quote was really intended and not a typo.

nvanheuverzwijn:
@PeterBrun I have highlighted details that seemed to me that would change the behavior but it might be OK. I just wanted to highlight them for you to make sure that everything is intended by this patch and not a typo or a bug in a search & replace.

PeterBrun:
@nvanheuverzwijn 👍

@luisdavim I have mostly tested on Windows. Using it there with some clusters. I have only tested the Linux build using the provided Linux tests (not with a real cluster).

luisdavim:
@sami-alajrami this LGTM but I wonder if there was any particular reason to start the helm and kubectl commands in a sub-shell instead of calling them directly.

luisdavim:
@mkubaczyk do you know what a sub-shell is being used?

luisdavim:
Hi can you take a look at the conflicts? Thanks.

sami-alajrami:
> @sami-alajrami this LGTM but I wonder if there was any particular reason to start the helm and kubectl commands in a sub-shell instead of calling them directly.

No, I don't think there was a special reason. Unless if it was to get error codes back .. but my memory is vague on it

luisdavim:
I that case I think we can merge this as soon as the conflicts are addressed. Thanks

PR #295, opened 2019-08-30
Adding support for SSM Params in the Helm values files

emilhdiaz:
This is a new feature implementation to add support for SSM parameter based variables. Both the helmsman state file and any values or secret files are supported. The solution is inspired by a similar feature in serverless-framework. 

With this feature a user can indicate that a particular value should be pulled from an SSM parameter at deploy time of the Helm Chart. 

For example, given a values.yaml file as such: 
```
ingress:
  enabled: true
  annotations:
     kubernetes.io/owner: {{ssm: /environment/owner}}
```
Helmsman will now perform a lookup of the SSM Param value indicated by the SSM param path above and inject the value into the values.yaml file. 

Encrypted SSM Params are also supported by appending `~true` to the SSM Param path. 

For example: 
```
ingress:
  enabled: true
  annotations:
     kubernetes.io/owner: {{ssm: /environment/owner~true}}
```

Environment variables within the SSM Param path are also supported. 

For example: 
```
ingress:
  enabled: true
  annotations:
     kubernetes.io/owner: {{ssm: /environment/${environment}/owner~true}}
```

The feature can be disabled globally with the `-no-ssm-subst` CLI flag or specifically only for values and secret files with the `-no-ssm-values-subst` CLI flag.

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
@@ -18,7 +19,8 @@ import (
 	"gopkg.in/yaml.v2"
 
 	"github.com/BurntSushi/toml"
-	"github.com/Praqma/helmsman/aws"
+	//"github.com/Praqma/helmsman/aws"
+	"helmsman/aws"
 	"github.com/Praqma/helmsman/azure"
 	"github.com/Praqma/helmsman/gcs"
 )
@@ -162,15 +164,17 @@ func substituteEnvInValuesFiles(s *state) {
 // Returns the path for the temp file
 func substituteEnvInYaml(file string) string {
 	rawYamlFile, err := ioutil.ReadFile(file)
-	var yamlFile string
+	yamlFile := string(rawYamlFile)
 	if err != nil {
 		logError(err.Error())
 	}
 	if !noEnvSubst {
-		yamlFile = substituteEnv(string(rawYamlFile))
-	} else {
-		yamlFile = string(rawYamlFile)
+		yamlFile = substituteEnv(yamlFile)
 	}
+	if !noSSMSubst {
+		yamlFile = substituteSSM(yamlFile)
+	}
+	log.Println("INFO: \n" + yamlFile)
 
 	dir, err := ioutil.TempDir(tempFilesDir, "tmp")
 	if err != nil {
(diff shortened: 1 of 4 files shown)


emilhdiaz:
@sami-alajrami Could use your help to clarify two things. 

CircleCI builds are failing because it seems the helmsman-test image is using an older version of Go prior to 1.12. Is there any intent to upgrade to 1.12 or should I make the code compatible with an older version? If so what is the minimum Go version Helmsman supports? 

I couldn't figure out how to import my changes to aws.go within my local environment without changing the import path. It seems that the `github.com/Praqma/helmsman/aws` import always pulls a version from github, which won't have my changes until this PR is merged. I had to resort to changing the import to a local one in order to build and test my changes. Is there a development workflow that I'm missing here?

sami-alajrami:
Hi @emilhdiaz thanks for working on this!
I have bumped the Go version in the test image to 1.12.9. 
For the importing the dependencies, please change it to use local path instead. There is no special workflow and you are right that it will just pull from github and ignore your changes. So it would be great if you fix it to use local path. Please let me know if you require more assistance with this.

emilhdiaz:
@sami-alajrami thanks for the clarifications! Should I convert all the paths to local paths in the PR? Or just the one for the aws.go file?

sami-alajrami:
@emilhdiaz please convert all (the aws and the google ones).

emilhdiaz:
@sami-alajrami I went ahead and converted all the local imports (aws, gcp, and azure) as you suggested. 

I also:
- added support for SSM params in the helmsman file itself
- improved the CLI flags for disabling SSM param substitution
- improved the logging around env and SSM variables substitution
- reorganized some of the code a bit

The CircleCI test is failing with a missing gcc exception, any ideas on what could be wrong with the issue?

utils.go
```diff
@@ -139,37 +145,40 @@ func toYAML(file string, s *state) {
...
 	if err != nil {
 		logError(err.Error())
 	}
-	if !noEnvSubst {
-		yamlFile = substituteEnv(string(rawYamlFile))
-	} else {
-		yamlFile = string(rawYamlFile)
+
+	yamlFile := string(rawYamlFile)
+	if !noEnvSubst && !noEnvValuesSubst {
```
luisdavim: I'm not sure this keeps the same functionality as before, is `substituteVarsInYaml` only used for the values files?

emilhdiaz: Hi @luisdavim, thanks for looking at my MR! 

To answer your question, yes, `substituteVarsInYaml` is only ever called from `substituteVarsInValuesFiles` and therefore only used for the values files. 

With the introduction of the `noSSMSubst` and `noSSMValuesSubst` flags, it was necessary to push the conditional checks down the stack to a point where flags for SSM param evaluation and Env param evaluation could be checked independently without one precluding the other. I also didn't want to create a second set of temporary files in order to add the SSM param feature, so this was my best choice (at the time) for a home for this logic. 

Do you see any issues with this approach?

emilhdiaz:
@sami-alajrami @luisdavim do you have any more feedback on this MR. Would love to get this merged in soon.

Issue #300, opened 2019-09-24
Feature Request: special ignored keys in yaml for using anchors in configuration


mkjmdski:
Hey!
I think this would be pretty simple to implement, so basically I would like to have a key in my yaml configuration which would look like that

```yaml
.service: &service-template
  enabled: true
  valuesFiles:
    - "../values/dev.yaml"
  secretsFiles:
    - "../values/dev.secrets.yaml"
  wait: true
  purge: true
```
so any key starting in yaml with . would be ignored, so later on I could write something like this

```yaml
  prometheus:
    <<: *service-template
    namespace: "prometheus"
    chart: "prometheus"
    version: "1.0.0"
    name: "prometheus"
    priority: -998
```

currently this configuration is producing error that `.service` is not  known. This would be very helpful in writing DRY configs in yaml for helmsman

mkubaczyk:
It's merged, you can wait for release or built yourself the newest version of the master branch

PR #302, opened 2019-10-01
Add appsTemplates omitempty YAML key to allow YAML anchors in state file

mkubaczyk:
The feature request #300 with ignored keys in YAML state file, where such keys can be used as YAML anchors to follow DRY rule is really valuable, so I thought of the simplest way to do this. Hence the PR I created with the least change I could do to achieve that.

### DIFF
--- a/docs/desired_state_specification.md
+++ b/docs/desired_state_specification.md
@@ -311,6 +311,52 @@ preconfiguredHelmRepos:
 
 > In this case you will manually need to execute `helm repo add myrepo1 <URL> --username= --password=`
 
+## AppsTemplates
+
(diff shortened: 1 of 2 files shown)


mkjmdski:
thanks!

mkubaczyk:
@sami-alajrami @luisdavim how you guys find it?

luisdavim:
have you tested it? as long as this doesn't break the `toml` DSF I'm fine with it.

mkubaczyk:
@luisdavim yes, it's working fine with yamls while at the same time not impacting tomls at all. Should I merge then?

PR #306, opened 2019-10-18
Dependencies upgrade

chenrui333:
Upgrade some dependencies:
- golang to v1.13.1
- helm, v2.15.0
- goreleaser v0.120.2
- k8s, v1.16.2


luisdavim:
@sami-alajrami I think we should probably update the pipeline and build images for a few different versions of kubernetes.

luisdavim:
@chenrui333 you should probably also update the dockerfile in the `test_files` folder.

chenrui333:
@luisdavim I think the `test_files` folder has been updated, right?

luisdavim:
I agree with @mkubaczyk probably not many people are using 1.16.x yet, helm v2.15 should be ok though. But if kubectl v1.16.x is compatible with k8s 1.14.x we should be ok.
This is why I made the comment about providing a few different versions of the Docker image with the last 2 or 3 versions of helm and kubectl. Something like latest and stable.

chenrui333:
I see, that is good point. I am using EKS for most of stuff, and my kubectl is up to date though.

chenrui333:
I can certainly update back to 1.14.x

luisdavim:
Yeah, it's probably better for now.

chenrui333:
Sounds good, just revert back to v1.14.8 :)

chenrui333:
Thanks @luisdavim!

PR #316, opened 2019-10-23
Limit bash usage

kpucynski:
Lower amount of wrapping - remove bash and call commands directly.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -11,7 +11,7 @@
     "internal/optional",
     "internal/trace",
     "internal/version",
-    "storage",
+    "storage"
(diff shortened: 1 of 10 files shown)


luisdavim:
this is already covered in https://github.com/Praqma/helmsman/pull/287 if we don't get an update from the author soon we can merge this instead.

luisdavim:
oh, he did update his PR...

kpucynski:
heh, I didn't saw this PR :D
I'll check.

kpucynski:
haha we did basically the same :D
lucky guy was faster ;)
closing.

PR #318, opened 2019-10-24
Add -group flag to limit execution of deployments to specific group of apps

mkubaczyk:
Since `-target` turned out to be quite useful, I'm proposing a variation of it in form of `-group` parameter. Each application can have one group it belongs to and the Helmsman's deployment can be not only limited to specific apps, which in bigger environments led to quite hard to prepare final commands, but also limited to groups of applications.
Thanks to this the control over apps deployment will stay on the same level it is for `-target` but with better visibility.

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -557,21 +557,3 @@ func concat(slices ...[]string) []string {
 	}
 	return slice
 }
-
-func isReleaseConsideredToRun(r *release) bool {
-	if len(targetMap) > 0 {
-		if _, ok := targetMap[r.Name]; ok {
-			return true
-		} else {
-			return false
-		}
-	}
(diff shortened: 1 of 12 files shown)


utils.go
```diff
@@ -557,3 +557,21 @@ func concat(slices ...[]string) []string {
 	}
 	return slice
 }
+
+func isReleaseConsideredToRun(r *release) bool {
```
luisdavim: nit: Maybe we could make this a method of the release object instead so we could do `r.IsConsideredToRun()`

mkubaczyk: I should have thought about that, applied the change, thanks. :-)

PR #320, opened 2019-10-24
Fixing some issues when running in tillerless mode

luisdavim:
One of the issues is that when the `helm-tiller` plugin is not initialised, tiller will be installed during the first use and that will break the expect json output.

```
12:58:55 ERROR: failed to unmarshal Helm CLI output: unexpected end of JSON input
 [] <nil> 0xc000352180 exit status 1 <nil> <nil> true [0xc0000d6000 0xc0000d6008 0xc000126218] [0xc000126218] [0xc000126210] [0x556fd0] 0xc000120720 <nil>}: exit status 1
stderr: 2019/10/24 12:58:55 ERROR: failed to unmarshal Helm CLI output: unexpected end of JSON input
```

The other issue I'm trying to address is that in some cases `helm search` displays a warning message before the table output and it ended up with the release version being set to `CHART VERSION`.

```
2019/10/24 13:25:05 Command returned with exit code: 1. And error message: Error: chart "my-chart" matching CHART VERSION not found in my-repo index. (try 'helm repo update'). improper constraint: CHART VERSION
Error: plugin "diff" exited with error
Error: plugin "tiller" exited with error
```

### DIFF
--- a/helm_helpers.go
+++ b/helm_helpers.go
@@ -334,19 +334,30 @@ func getChartVersion(r *release) (string, string) {
 		Description: "getting latest chart version " + r.Chart + "-" + r.Version + "",
 	}
 
-	if exitCode, result := cmd.exec(debug, verbose); exitCode != 0 || strings.Contains(result, "No results found") {
+	var (
+		exitCode int
+		result   string
+	)
+
+	if exitCode, result = cmd.exec(debug, verbose); exitCode != 0 || strings.Contains(result, "No results found") {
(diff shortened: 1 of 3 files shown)


helm_helpers.go
```diff
@@ -520,6 +531,21 @@ func initHelmClientOnly() (bool, string) {
 	return true, ""
 }
 
+// initHelmTiller initializes the helm tiller plugin
+func initHelmTiller() (bool, string) {
+	cmd := command{
+		Cmd:         "helm",
+		Args:        []string{"tiler", "install"},
```
jmontroy90: Should be `tiller`?

PR #322, opened 2019-10-25
fix tests for go 1.13

luisdavim:
Tests were failing for me after upgrading to the latest version of go, this fixes them.

### DIFF
--- a/init_test.go
+++ b/init_test.go
@@ -2,6 +2,11 @@ package main
 
 import "testing"
 
+var _ = func() bool {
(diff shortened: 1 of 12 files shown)


init_test.go
```diff
@@ -2,6 +2,11 @@ package main
 
 import "testing"
 
+var _ = func() bool {
+	testing.Init()
+	return true
+}()
+
```
luisdavim: This is the actual fix the rest is just `go ftm`

PR #337, opened 2019-12-11
use the current helm state to identify untracked releases

### LINKED ISSUE #336: Helmsman is not deleting untracked releases
Looks like when listing the existing releasees, helmsman only looks at the ones that are also present in the desired state so it never finds any untracked releases.

luisdavim:
This should fix #336. 
The code could use some more cleaning but I'm trying to keep the changes to the minimum to stay out of the way of @mkubaczyk's HelmV3 PR. I think we should release this before the Helm v3 rewrite so we have a working Helm v2 version.

### DIFF
--- a/helm_helpers.go
+++ b/helm_helpers.go
@@ -24,6 +24,7 @@ type releaseState struct {
 	Updated         time.Time
 	Status          string
 	Chart           string
+	Name            string
 	Namespace       string
 	TillerNamespace string
 }
@@ -198,6 +199,7 @@ func buildState() {
 			Updated:         time,
 			Status:          rel.Releases[i].Status,
 			Chart:           rel.Releases[i].Chart,
+			Name:            rel.Releases[i].Name,
 			Namespace:       rel.Releases[i].Namespace,
 			TillerNamespace: rel.Releases[i].TillerNamespace,
 		}
@@ -227,7 +229,6 @@ func helmReleaseExists(r *release, status string) (bool, releaseState) {
 
 // getReleaseRevision returns the revision number for a release
 func getReleaseRevision(rs releaseState) string {
-
 	return strconv.Itoa(rs.Revision)
 }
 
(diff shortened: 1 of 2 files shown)


kube_helpers.go
```diff
@@ -480,12 +480,14 @@ func getHelmsmanReleases() map[string]map[*release]bool {
 				continue
```
mkubaczyk: How about using the approach I took on https://github.com/Praqma/helmsman/blob/738d7d6905616d5e5e71a93e422e31c33ba955fa/kube_helpers.go#L398 ? It simplifies parsing and while I completely understand the need of fixing this asap before helmsman 3 goes on, this would help us merge it in the future :-) If you follow the `-o name`, then rest of changes I did while migrating to helm 3 simplifies the code of this whole function

luisdavim: Sure, I'll update the pr. Thanks

luisdavim: I've updated the PR to follow your approach.

helm_helpers.go
```diff
@@ -614,7 +619,7 @@ func cleanUntrackedReleases() {
 	} else {
 		for ns, releases := range toDelete {
 			for r := range releases {
-				if r.isReleaseConsideredToRun() {
+				if !r.isReleaseConsideredToRun() {
```
jmontroy90: Why this inversion? AFAIK this says "did you specifically tell me to manage this release via startup flags?". So we're now saying "if we don't manage this release, don't delete it", and below, "otherwise, the flags were populated with this release, we do manage this release -- delete it". Doesn't that mean this will only delete releases we explicitly include on startup in `target` or `group`?

luisdavim: this was another bug, `r.isReleaseConsideredToRun()` basically means "is this release being managed" and if this condition is true we skip the release, in this case we want to skip the release if it's not being managed.

kube_helpers.go
```diff
@@ -475,18 +476,18 @@ func getHelmsmanReleases() map[string]map[*release]bool {
...
-				fields := strings.Fields(lines[i])
-				if _, ok := releases[ns]; !ok {
-					releases[ns] = make(map[*release]bool)
-				}
-				for _, r := range s.Apps {
-					if r.Name == fields[0][0:strings.LastIndex(fields[0], ".v")] {
-						releases[ns][r] = true
-					}
+			}
+			r = regexp.MustCompile(`(^\w+\/|\.v\d+$)`).ReplaceAllString(r, "")
```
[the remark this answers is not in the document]
luisdavim: with `-o name` the returned list of resources is in the format `<resourceType>/<resoureceName>.v<revNumber>`.

helm_helpers.go
jmontroy90: Got it. Thanks.

kube_helpers.go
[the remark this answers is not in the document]
jmontroy90: Super, thanks.

mkubaczyk:
@luisdavim since your fix is merged, can we start pushing current master as legacy branch and merge helm 3 version slowly to master?

PR #340, opened 2019-12-13
add resourcequota

ivelichkovich:
command.exec was not working with "helm init --override" commands, even the existing one didn't work.

Will add documentation soon just opening up PR for code review for now.

### DIFF
--- a/docs/how_to/README.md
+++ b/docs/how_to/README.md
@@ -13,6 +13,7 @@ This page contains a list of guides on how to use Helmsman.
     - [Label namespaces](namespaces/labels_and_annotations.md)
     - [Set resource limits for namespaces](namespaces/limits.md)
     - [Protecting namespaces](namespaces/protection.md)
+    - [Namespace resource quotas](namespaces/quotas.md)
 - Deploying Helm Tiller
     - [Using existing Tillers](tiller/existing.md)
     - [Deploy shared Tiller in kube-system](tiller/shared.md)
--- a/docs/how_to/namespaces/quotas.md
+++ b/docs/how_to/namespaces/quotas.md
@@ -1,0 +1,45 @@
+---
+version: ???
+---
+
+# Define resource quotas for namespaces
+
+You can define namespaces to be used in your cluster. If they don't exist, Helmsman will create them for you. You can also define how much resource limits to set for each namespace.
+
+You can read more about the `Quotas` specification [here](https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/quota-memory-cpu-namespace/#create-a-resourcequota).
+
+```toml
+#...
+[namespaces]
+  [namespaces.helmsman1]
+    [namespaces.helmsman1.quotas]
+    pods = "25"
+      [namespaces.helmsman1.quotas.requests]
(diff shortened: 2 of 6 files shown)


luisdavim:
Thanks @ivelichkovich can you also update the tests so they don't fail?

command.go
```diff
@@ -41,14 +41,16 @@ func (c command) exec(debug bool, verbose bool) (int, string, string) {
...
 	if debug {
 		log.Println("INFO: " + c.Description)
 	}
 	if verbose {
-		log.Println("VERBOSE: " + c.Cmd + " " + strings.Join(args, " "))
+		log.Println("VERBOSE: " + command)
 	}
 
-	cmd := exec.Command(c.Cmd, args...)
+	cmd := exec.Command("sh", "-c", command)
```
luisdavim: why is this needed? what error where you getting when calling the command directly?

ivelichkovich:
Will add documentation soon

command.go
ivelichkovich: the override flags were not working because regular exec.command ignores quotes in args.
so this existing override  was never passed to tiller 
[]string{"--override", "'spec.template.spec.containers[0].command'='{/tiller,--storage=secret}'"}
and my overrides were not working as well

luisdavim: Ah, I see. In that case the solution should be to include the quotes in the string, something like:
```go
[]string{"--override", `"'spec.template.spec.containers[0].command'='{/tiller,--storage=secret}'"`}
```
What do you think?

ivelichkovich: Including quotes did not work, for example the command you posted was not being processed, instead of globally changing it to "sh -c" though I changed just that command to run as "sh -c" (got that idea from the command_test.go). Otherwise "sh -c" was breaking all the tests for directories with white space (that we may want to consider not supporting but doesn't matter too much to me). I updated to add documentation as well.

luisdavim: the line I wrote was just an example, how if the final command supposed to look like, with quotes? There must be a solution, the problem with using `sh -c` is that it won't work for windows users...

ivelichkovich: Hmm not sure, will need to think about it, try it out though maybe I was incorrect but the 'spec.template.spec.containers[0].command'='{/tiller,--storage=secret}' never ends up getting set on your tiller in this current version

ivelichkovich: I think go isn't honoring the single quotes around the different pieces of the --override and helm requires those otherwise it ignores the override

mkubaczyk:
@ivelichkovich you might need to change your MR's base branch, since master is currently reserved for Helmsman v3. Point it at branch 1.x

luisdavim:
@ivelichkovich maybe if you split this into 2 PRs we can merge the tiller resources part and we can look into the tiller setup issue in a separate PR. I've changed the base for this PR to the `1.x` branch since master is now tracking the Helm v3 version of helmsman where tiller is no longer needed.

ivelichkovich:
Okay we can do just resource quotas, we're going to be using helm3 with this so can we point it back to master and I'll rebase? with helm3 obviously tiller resources won't be needed so removed that part

PR #342, opened 2019-12-20
Fix/untracked releases for helmsman 3.0.0-beta

sami-alajrami:
this PR updates the detection of untracked releases to match helm3 and the recent helmsman rewrites.
The expected behaviour remains the same. If a release is no longer in the desired state file(s) but is deployed in the cluster and labeled as MANAGED-BY=HELMSMAN, then it is considered untracked and should be removed (unless user enables the `--keep-untracked-releases` flag)

@mkubaczyk  I removed the check if the release is targeted before deleting it because it didn't make sense that the release is not in the desired state but is being targeted. Maybe I am missing something?

### DIFF
--- a/examples/example.yaml
+++ b/examples/example.yaml
@@ -1,4 +1,4 @@
-# version: v1.5.0
+# version: v3.0.0-beta1
 # metadata -- add as many key/value pairs as you want
 metadata:
   org: "example.com/$ORG_PATH/"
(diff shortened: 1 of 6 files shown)


mkubaczyk:
@sami-alajrami some people can have more than 1 DSF to describe their clusters and limit the execution with target flag hence such approach

sami-alajrami:
@mkubaczyk for that use case, I am working on adding a `scope` or something similar to the DSF so that the releases are labeled with that scope and can be distinguished by Helmsman. This way, Helmsman can then operate only on releases deployed with the supplied DSF(s). 
I plan to add this later today in a separate PR before we release v3.0.0-beta1.

PR #347, opened 2019-12-22
Add goroutines-based parallel execution of plan and charts availability checking

mkubaczyk:
Since plan (aka. diffing) is on a second place of time execution right after charts installation/upgrade, I'm adding goroutines to diffing and checking charts availability to improve this part a bit.
Fixed tests and checked that locally, but I'd like someone else to confirm it's ok before it lands in master.

### DIFF
--- a/internal/app/helm_helpers.go
+++ b/internal/app/helm_helpers.go
@@ -2,6 +2,7 @@ package app
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/url"
 	"os"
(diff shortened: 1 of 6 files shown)


internal/app/helm_helpers.go
```diff
@@ -160,56 +162,72 @@ func getReleaseChartVersion(rs releaseState) string {
...
-		}
-		if validateCurrentChart {
-			if isLocalChart(r.Chart) {
-				cmd := command{
-					Cmd:         helmBin,
-					Args:        []string{"inspect", "chart", r.Chart},
-					Description: "Validating [ " + r.Chart + " ] chart's availability",
-				}
+		wg.Add(1)
+		go func(app string, r *release, wg *sync.WaitGroup, c chan string) {
```
luisdavim: nit: moving this into a separate function would be cleaner in my opinion.

PR #349, opened 2019-12-24
reorganisig the code using receiver methods

luisdavim:
* Reorganised the code using receiver methods
* Merged `releaseInfo` and `releaseState` structs into one
* Removed the `godep` files
* Updated the Makefile to use go modules
* Cleaned up trailing white spaces (use `Hide whitespace changes` for a better review experience)


luisdavim:
I still think there's some cleaning to be done but I'm ok with getting this reviewed and merged now and taking another pass at it latter in a separate PR.

PR #353, opened 2019-12-30
another round of code cleaning

luisdavim:


internal/app/decision_maker.go
```diff
@@ -149,37 +153,46 @@ var releaseNameExtractor = regexp.MustCompile(`sh\.helm\.release\.v\d+\.`)
 // getHelmsmanReleases returns a map of all releases that are labeled with "MANAGED-BY=HELMSMAN"
 // The releases are categorized by the namespaces in which they are deployed
 // The returned map format is: map[<namespace>:map[<helmRelease>:true]]
-func (cs currentState) getHelmsmanReleases() map[string]map[string]bool {
+func (cs *currentState) getHelmsmanReleases(s *state) map[string]map[string]bool {
 	var lines []string
+	const outputFmt = "custom-columns=NAME:.metadata.name,NS:.metadata.namespace,CTX:.metadata.labels.HELMSMAN_CONTEXT"
```
luisdavim: these changes should fix #355 and make things faster since only one `kubectl` command is issued to get all releases.

internal/app/plan.go
```diff
@@ -36,24 +37,25 @@ type orderedCommand struct {
 
 // plan type representing the plan of actions to make the desired state come true.
 type plan struct {
+	sync.Mutex
```
luisdavim: adding this here and to the `currentState` `type` should fix #354

internal/app/plan.go
```diff
@@ -63,8 +65,17 @@ func (p *plan) addCommand(cmd command, priority int, r *release) {
 	p.Commands = append(p.Commands, oc)
 }
 
+// logDecision adds the decisions made to the plan.
+// Depending on the debug flag being set or not, it will either log the the decision to output or not.
+func (p *plan) logDecision(decision string, priority int, decisionType decisionType) {
+	p.addDecision(decision, priority, decisionType)
+	log.Debug(decision)
```
sami-alajrami: same as the comment above .. this logs the plan decisions with DEBUG prefix when only `verbose` is used. In fact those decisions are logged anyway in the plan in all cases. So maybe this log can be deleted here.

internal/app/command.go
```diff
@@ -31,42 +36,51 @@ func (c *command) exec(debug bool, verbose bool) (int, string, string) {
 	}
 
 	log.Verbose(c.Description)
+	log.Debug(c.String())
```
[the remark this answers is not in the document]
luisdavim: I've added the `if flags.debug {}` condition on the logger side.

internal/app/plan.go
luisdavim: I've removed the whole function and use `addDecision` directly everywhere.

PR #364, opened 2020-01-03
stop adding stable helm repos by default

### LINKED ISSUE #359: should we disable adding stable/incubator helm repos by default?
Helm 3 [no longer sets repos by default](https://helm.sh/docs/topics/v2_v3_migration/#overview-of-helm-3-changes):

```
local or stable repositories are not set-up by default
```
Helmsman v3.0.0-betaX still adds those repos by default and we have  a flag to disable that behaviour. Should we reverse this (not add repos by default and provide a flag to add them) to be aligned with the helm 3 behaviour? Or the current behaviour is good enough? 

Any thoughts @luisdavim @mkubaczyk

sami-alajrami:
Fixes #359

### DIFF
--- a/examples/example.toml
+++ b/examples/example.toml
@@ -67,10 +67,10 @@ context= "test-infra"   # defaults to "default" if not provided
   argo = "https://argoproj.github.io/argo-helm"
   jfrog = "https://charts.jfrog.io"
   # stable = "https://kubernetes-charts.storage.googleapis.com"
-	# incubator = "http://storage.googleapis.com/kubernetes-charts-incubator"
-#  myS3repo = "s3://my-S3-private-repo/charts"
-#  myGCSrepo = "gs://my-GCS-private-repo/charts"
-#  custom = "https://user:pass@mycustomrepo.org"
+  # incubator = "http://storage.googleapis.com/kubernetes-charts-incubator" 
+  # myS3repo = "s3://my-S3-private-repo/charts"
+  # myGCSrepo = "gs://my-GCS-private-repo/charts"
+  # custom = "https://user:pass@mycustomrepo.org"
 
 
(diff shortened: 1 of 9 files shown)


examples/example.toml
```diff
@@ -66,6 +66,8 @@ context= "test-infra"   # defaults to "default" if not provided
 [helmRepos]
   argo = "https://argoproj.github.io/argo-helm"
   jfrog = "https://charts.jfrog.io"
+  # stable = "https://kubernetes-charts.storage.googleapis.com"
+	# incubator = "http://storage.googleapis.com/kubernetes-charts-incubator"
```
luisdavim: nit: why not the same indentation?

internal/app/utils.go
```diff
@@ -57,7 +57,7 @@ func fromTOML(file string, s *state) (bool, string) {
 	if _, err := toml.Decode(tomlFile, s); err != nil {
 		return false, err.Error()
 	}
-	addDefaultHelmRepos(s)
+	//addDefaultHelmRepos(s)
```
luisdavim: nit: if the function was removed there's no point in keeping this commented out here.

internal/app/utils.go
```diff
@@ -109,7 +109,7 @@ func fromYAML(file string, s *state) (bool, string) {
 	if err = yaml.UnmarshalStrict([]byte(yamlFile), s); err != nil {
 		return false, err.Error()
 	}
-	addDefaultHelmRepos(s)
+	//addDefaultHelmRepos(s)
```
luisdavim: nit: if the function was removed there's no point in keeping this commented out here.

examples/example.toml
sami-alajrami: copy-paste error :)

Issue #391, opened 2020-01-21
helmsman parallelization needs tuning or maybe I am missing something (2000 threads created at apply)


sharkymcdongles:
I have a deployment plan yaml with 180 or so apps in it. When I run apply against the plan it starts to do a helm search during the validation step. 

At this stage I see the cpu threads spike to 2000 and then it dies and fails with a chart cannot be found error despite it existing in the target chart repo. It seems to me it fails with this because the thread dies without a 200 from chartmuseum due to the cpu spike causing the check to fail.

Anyways attached is a picture of this happening. Would be really nice to know if there is a way to have it only run 5 or so threads in parallel.

Cheers! I will try and dive into the code and see what I can find as well.

https://i.imgur.com/cdkgJyK.png

Helm Version: 3.0.2
Helmsman Version: 3.0.1
kubernetes: 1.16.0
arch: amd64

sharkymcdongles:
I found this bit in the code:

`func validateReleaseCharts(s *state) error {
        var fail bool
        wg := sync.WaitGroup{}
        c := make(chan string, len(s.Apps))
        for app, r := range s.Apps {
                wg.Add(1)
                go r.validateChart(app, s, &wg, c)
        }
        wg.Wait()
        close(c)
        for err := range c {
                if err != "" {
                        fail = true
                        log.Error(err)
                }
        }
        if fail {
                return errors.New("chart validation failed")
        }
`

This needs something to limit it I think. I added a sleep here and it helped get to the next step where the makePlan func caused the same problem. Will try adding sleeps on all go routines I encounter where this happens just to see if I can get past this issue.

The actual clean solution would be to do something like this: https://github.com/korovkin/limiter

mkubaczyk:
That is indeed an issue when such amount of apps is defined in DSF. I’ll find myself a time soon to implement concurrency limit on that

luisdavim:
I've opened a PR that should fix this issue, can you please give it a try?
Thanks.

sharkymcdongles:
@luisdavim thanks for the hard work. I have built the binary from your PR and now it works. Well done! I know my PR was bad, but https://meta.m.wikimedia.org/wiki/Cunningham%27s_Law

:+1:

PR #394, opened 2020-01-21
sleep hack

sharkymcdongles:
This is one dumb way to quickly fix the fact that larger deployments with over 40 apps will have issues and crash the apply step.

https://github.com/Praqma/helmsman/issues/391

### DIFF
--- a/internal/app/release.go
+++ b/internal/app/release.go
@@ -10,6 +10,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"time"
 )
 
 // release type representing Helm releases which are described in the desired state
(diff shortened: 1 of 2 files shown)


luisdavim:
how does this solve the problem?

internal/app/release.go
```diff
@@ -184,8 +186,8 @@ func (r *release) validateChart(app string, s *state, wg *sync.WaitGroup, c chan
 			}
 			matches := versionExtractor.FindStringSubmatch(result.output)
 			if len(matches) == 2 {
-				version := strings.Trim(matches[1], `'"`)
-				if strings.Trim(r.Version, `'"`) != version {
+				version := matches[1]
```
luisdavim: why are you reverting this?

luisdavim:
This is not really the solution for the problem.

sharkymcdongles:
I know it doesn't solve it and said as much. Is a hack until I have time to
properly fix it or a maintainer does.

luisdavim:
I'll open a pr to solve this in a few minutes.

PR #395, opened 2020-01-22
limit the number of parallel go routines, fixes #391

### LINKED ISSUE #391: helmsman parallelization needs tuning or maybe I am missing something (2000 threads created at apply)

luisdavim:


luisdavim:
Sem is using empty structs because they have no state and use less memory than a int or even a book that only has 2 states.

Implementing semaphores using channels is simple enough and if something is simple to implement it's better to do it than to add a new dependency...

mkubaczyk:
@luisdavim I'm more than pleased with such explanation. Thanks! :-)

PR #405, opened 2020-01-30
Enhance target flag to ignore all releases states not targeted 

mkubaczyk:
This should shorten time needed for targeted deployments as it does not waste time for checking namespaces, charts availability and current state of releases not included in -target flag.

### DIFF
--- a/internal/app/kube_helpers.go
+++ b/internal/app/kube_helpers.go
@@ -13,7 +13,7 @@ import (
 // addNamespaces creates a set of namespaces in your k8s cluster.
 // If a namespace with the same name exists, it will skip it.
 // If --ns-override flag is used, it only creates the provided namespace in that flag
-func addNamespaces(s state) {
+func addNamespaces(s *state) {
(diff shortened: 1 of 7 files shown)


luisdavim:
@mkubaczyk won't this result in untracked releases never being cleaned up?

internal/app/kube_helpers.go
```diff
@@ -13,8 +13,14 @@ import (
 // addNamespaces creates a set of namespaces in your k8s cluster.
 // If a namespace with the same name exists, it will skip it.
 // If --ns-override flag is used, it only creates the provided namespace in that flag
-func addNamespaces(namespaces map[string]namespace) {
+func addNamespaces(s state) {
```
luisdavim: Since we use pointers to `state` everywhere perhaps we should use a pointer here as well.

mkubaczyk: good catch, thanks :-) done

PR #409, opened 2020-02-04
Make decide exiting on pending-upgrade chart state

mkubaczyk:
This may be specially useful on CI, when some jobs may cause racing condition


luisdavim:
Unrelated to your changes but looking at this makes me think we should probably have constants for the different possible release states and use them here.

mkubaczyk:
@luisdavim sounds ok, let me know if that's something you've been thinking of

PR #410, opened 2020-02-05
Changes to standard output and Slack messages

sharkymcdongles:
This adds version to the standard output to show what version is being installed/upgraded.

This improves slack output to not use pretext to prevent it from being cut off.


sharkymcdongles:
I think this is to a point where it can be reviewed.

internal/app/utils.go
```diff
@@ -374,22 +374,33 @@ func deleteFile(path string) {
...
 		color = "#FF0000" // red
 	}
 
+	var contentBold string
 	var pretext string
+
 	if content == "" {
-		pretext = "No actions to perform!"
+		pretext = "*No actions to perform!*"
+		contentBold = content
```
luisdavim: Nti: At this point both variables are empty strings,  this line seems redundant.

luisdavim:
Why make everything bold?

sharkymcdongles:
> Why make everything bold?

The original code used the pretext field for displaying the plan, and pretext is bold by design. I just assumed you'd want to mirror the same bolding that was there. I can change it to whatever you guys think is best. I just extended what was there with no real plan except mirroring the original functionality best I could while adding some stuff. Please let me know what formatting you prefer.

As for the redundant bit, I will commit changes tomorrow to remedy that mistake on my part.

Thanks for the reply!

internal/app/utils.go
sharkymcdongles: I went ahead and removed it as suggested.

Issue #400, opened 2020-01-25
Local charts installation is not idempotent


evgkrsk:
Plan is always contain uninstalling and reinstalling local chart for me:
```
evg at latitude in ~ $helmsman -no-fancy -apply -f Helmsman.yaml 
2020-01-25 20:06:17 INFO: Parsed YAML [[ Helmsman.yaml ]] successfully and found [ 1 ] apps
2020-01-25 20:06:17 INFO: Validating desired state definition...
2020-01-25 20:06:17 INFO: Setting up kubectl...
2020-01-25 20:06:17 INFO: Setting up helm...
2020-01-25 20:06:17 INFO: Setting up namespaces...
2020-01-25 20:06:17 INFO: Validating charts...
2020-01-25 20:06:17 INFO: Preparing plan...
2020-01-25 20:06:17 INFO: Acquiring current Helm state from cluster...
2020-01-25 20:06:18 INFO: Checking if any Helmsman managed releases are no longer tracked by your desired state ...
2020-01-25 20:06:18 INFO: No untracked releases found
2020-01-25 20:06:18 NOTICE: -------- PLAN starts here --------------
2020-01-25 20:06:18 NOTICE: Release [ logger ] is desired to use a new chart [ /home/evg/git/k8s-event-logger/chart ]. Delete of the current release will be planned and new chart will be installed in namespace [ default ] -- priority: 0
2020-01-25 20:06:18 NOTICE: -------- PLAN ends here --------------
2020-01-25 20:06:18 INFO: Executing plan... 
2020-01-25 20:06:18 NOTICE: Deleting release [ logger ] in namespace [ default ]
2020-01-25 20:06:19 NOTICE: release "logger" uninstalled

2020-01-25 20:06:19 NOTICE: Finished: Deleting release [ logger ] in namespace [ default ]
2020-01-25 20:06:19 NOTICE: Installing release [ logger ] in namespace [ default ]
2020-01-25 20:06:20 NOTICE: NAME: logger
LAST DEPLOYED: Sat Jan 25 20:06:19 2020
NAMESPACE: default
STATUS: deployed
REVISION: 1
TEST SUITE: None
NOTES:
To verify that the k8s-event-logger pod has started, run:

  kubectl --namespace=default get pods -l "app.kubernetes.io/name=k8s-event-logger,app.kubernetes.io/instance=logger"

2020-01-25 20:06:20 NOTICE: Finished: Installing release [ logger ] in namespace [ default ]
[...]

PR #411, opened 2020-02-11
Fix proposal for issue #400

Surian:
Hello :)

This is a fix proposal for issue #400 

Chart name in DSF file can be a chart from a repository, a path to a local directory or a TGZ archive.
DSF file is currently not idempotent when using a path to a TGZ archive or a path to a local directory where the directory name is different from the chart name because it rely on a regex on the chart name field to get the chart name.

When the chart name is a path to a TGZ archive, the extracted chart name will always be wrong. When it's a local directory, the chart name will be wrong if the directory name is differont from the chart name defined in Chart.yaml.

This PR is a proposal to fix these two issues by executing "helm show chart" and getting chart name from the command output instead of using a regex on the provided chart name in DSF file. This way, the extracted chart name is correct in all cases.

### DIFF
--- a/internal/app/helm_helpers.go
+++ b/internal/app/helm_helpers.go
@@ -37,7 +37,7 @@ func extractChartName(releaseChart string) string {
 	for _, v := range strings.Split(result.output, "\n") {
 		split := strings.Split(v, ":")
 		if len(split) == 2 && split[0] == "name" {
-			name = strings.TrimSpace(split[1])
+			name = strings.Trim(split[1], `"' `)
 			break
 		}
 	}
--- a/internal/app/helm_helpers.go
+++ b/internal/app/helm_helpers.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"net/url"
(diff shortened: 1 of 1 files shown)


internal/app/helm_helpers.go
```diff
@@ -25,18 +24,25 @@ func helmCmd(args []string, desc string) command {
...
+	}
 
-	m := chartNameExtractor.FindStringSubmatch(releaseChart)
-	if len(m) == 2 {
-		return m[1]
+	name := ""
+	for _, v := range strings.Split(result.output, "\n") {
+		split := strings.Split(v, ":")
+		if len(split) == 2 && split[0] == "name" {
+			name = strings.TrimSpace(split[1])
```
luisdavim: I think:
```go
name = strings.Trim(split[1], `"' `)
```
would be better here.

Surian: Indeed, this should be more reliable.

Surian:
@luisdavim change made!

PR #421, opened 2020-03-20
Introducing helmsman lifecycle hooks

### LINKED ISSUE #360: Feature suggestion: have the ability to add pre and post deploy actions
The idea I have is to allow referencing k8s yaml manifests to be applied before or after operating on a release in the DSF. 

This is useful in scenarios where you need to deploy something that's not part of a 3rd party helm chart. For example, install CRDs (if they are not packaged in the chart), create a secret or a configmap needed by the chart, or perform some post-installation processing.

Some of these cases can be implemented by helm lifecycle hooks if you manage the chart. But with 3rd party charts, you don't always get these.

The format could be something like this:
```yaml
...
nexus:
      pre-install: create-my-nexus-password-secret.yaml
      namespace: utils
      enabled: true
      chart: "stable/sonatype-nexus"
      valuesFile: "values-files/nexus-values.yaml"
      version: "1.21.3"
      priority: -5
```

Upon installation/upgrade of this release, we can do `kubectl apply` and upon deletion, we do `kubectl delete`. 
I would not want to allow arbitrary scripts for these hooks, because we can't control their effect (example we can't easily reverse the effect of a script if the release is deleted).
If this feature is desired, I can have a go at implementing it.

sami-alajrami:
this PR introduces helmsman life-cycle hooks which allow applying certain k8s manifests before and/or after helmsman operations. It fixes #360 
Along the way, this PR allows all files referred to in the DSF  (values files, secret files, lifecycle hooks, etc) to be fetched from URLs (e.g. another public git repo), cloud buckets or local file paths.


eperdeme:
Great work. I just went to install cert-manager from Jetstack and was "ahhh CRDs not bundled with Helm"....then saw an email with this pull request which magically fixes my new problem today.

Thanks !

sami-alajrami:
@eperdeme great to hear. Would you be able to test this PR?

examples/example.toml
```diff
@@ -41,23 +46,23 @@ context= "test-infra"   # defaults to "default" if not provided
...
-        cpu = "300m"
-        memory = "200Mi"
-      [namespaces.production.limits.defaultRequest]
-        cpu = "200m"
-        memory = "100Mi"
-    [[namespaces.production.limits]]
-      type = "Pod"
-      [namespaces.production.limits.max]
-        memory = "300Mi"
+  #   protected = true
```
luisdavim: Why comment this out?

examples/example.yaml
```diff
@@ -30,26 +30,30 @@ settings:
...
-        default:
-          cpu: "300m"
-          memory: "200Mi"
-        defaultRequest:
-          cpu: "200m"
-          memory: "100Mi"
-      - type: Pod
-        max:
-          memory: "300Mi"
+    # protected: true
```
luisdavim: Why comment this out?

internal/app/utils.go
```diff
@@ -139,15 +138,55 @@ func toYAML(file string, s *state) {
...
+func substituteVarsInStaticFiles(s *state) {
 	for _, v := range s.Apps {
 		if v.ValuesFile != "" {
 			v.ValuesFile = substituteVarsInYaml(v.ValuesFile)
 		}
 		if v.SecretsFile != "" {
 			v.SecretsFile = substituteVarsInYaml(v.SecretsFile)
 		}
+		if (hooks{}) != v.Hooks {
+			if v.Hooks.PreInstall != "" {
```
luisdavim: maybe if we convert the hooks structure into a map here we could make this code less verbose.

internal/app/release.go
```diff
@@ -34,6 +34,20 @@ type release struct {
 	HelmFlags    []string          `yaml:"helmFlags"`
 	NoHooks      bool              `yaml:"noHooks"`
 	Timeout      int               `yaml:"timeout"`
+	Hooks        hooks             `yaml:"hooks"`
+}
+
+// hooks type which defines life-cycle hooks
+type hooks struct {
```
luisdavim: I think this struct is fine for reading the data from the helmsman file but during the validation we could convert it into another one where we keep the "hook actions" in a `map[string]string` that way we can cleanup the code that needs to loop through them, what do you think?

orefalo:
Hi, can't wait for this to get released. I am doing everything manually now... 

If you had a compiled bin for osx, I can ever test and report any issues!

Cheers

sami-alajrami:
I am picking it up now to resolve the reviews above and then will merge it.

mkubaczyk:
You can also change the release version in docs to 3.4.0 :-)

internal/app/release.go
sami-alajrami: Thanks to the help of @zanderhavgaard , the struct is now replaced with a map.

PR #425, opened 2020-03-28
 Add DRY-ed Example 

Xtigyro:


examples/dried-example/README.md
```diff
@@ -0,0 +1,9 @@
+# DRY-ed Example
+
+## Execution Plan
+
+To deploy the DRY-ed example - please run:
+
+```bash
+helmsman -apply -no-default-repos -f config/helmsman.yaml
```
sami-alajrami: `--no-default-repos` is deprecated in helmsman v3.X

Xtigyro:
> 
> 
> thanks @Xtigyro for providing this missing example.
> I would rather name the readme and the directories as AppsTemplates than `dried`. Additionally, it would be nice to explain in the readme that you would use the appsTemplates to use named templates and achieve DRY. It is worth mentioning also, that this would work on YAML desired state files and not with TOML (unless if I am wrong).
> Finally, it would be great to link this readme/example in this docs landing page (under Misc) to make it easier to find https://github.com/Praqma/helmsman/tree/master/docs/how_to

Thanks for the so detailed and nice feedback, @sami-alajrami!
Will work on it and let ya know (when ready).

Xtigyro:
@sami-alajrami Hey - do you think it's ready now?

PR #430, opened 2020-04-03
Use --upgrade instead of --force-upgrade with helm init

pmichna:
`helm init --force-upgrade` does not work without `--upgrade` option.

### DIFF
--- a/helm_helpers.go
+++ b/helm_helpers.go
@@ -507,7 +507,7 @@ func deployTiller(namespace string, serviceAccount string, defaultServiceAccount
 	}
 	cmd := command{
 		Cmd:         "helm",
-		Args:        concat([]string{"init", "--force-upgrade"}, maxHistory, sa, tillerNameSpace, tls, storageBackend),
+		Args:        concat([]string{"init", "--upgrade", "--force-upgrade", "--wait"}, maxHistory, sa, tillerNameSpace, tls, storageBackend),
 		Description: "initializing helm on the current context and upgrading Tiller on namespace [ " + namespace + " ].",
 	}
 


helm_helpers.go
```diff
@@ -507,7 +507,7 @@ func deployTiller(namespace string, serviceAccount string, defaultServiceAccount
 	}
 	cmd := command{
 		Cmd:         "helm",
-		Args:        concat([]string{"init", "--force-upgrade"}, maxHistory, sa, tillerNameSpace, tls, storageBackend),
+		Args:        concat([]string{"init", "--upgrade"}, maxHistory, sa, tillerNameSpace, tls, storageBackend),
```
bacongobbler, line 510: I assume the intention is `helm init --upgrade --force-upgrade`, in which case the PR should just add `--upgrade` to the arg list.

pmichna:
Thanks @luisdavim for merging! Are you going to make a new release with the changes soon?

luisdavim:
I've pushed a new tag from the `1.x` branch, not sure if there's anything else needed to create a release for the 1.x version.
@sami-alajrami ?

sami-alajrami:
@luisdavim the release job was triggered but failed to build [here](https://circleci.com/gh/Praqma/helmsman/1701?pipelines-ui-opt-out) because of missing dependencies. The [release job](https://github.com/Praqma/helmsman/blob/91a7d3148c3af0a01365899ead77c69f41d98b4e/.circleci/config.yml#L40) probably needs to be fixed.

luisdavim:
I've created this https://github.com/Praqma/helmsman/pull/433 to see if it fixes the issue, I think the problem was that at some point we had our dependencies vendored and now we don't so we need to fetch them.

Issue #361, opened 2020-01-02
Parallelise releases with the same priority


luisdavim:
In Helmsman v3 we already parallelised some actions to make the execution faster, should we also add some parallelisation to the plan execution? We can group the commands in the plan by their priority and execute all the commands with the same priority in parallel.

PR #431, opened 2020-04-06
feat: Parallelise releases with the same priority

luisdavim:
fixes: #361

### DIFF
--- a/internal/app/plan.go
+++ b/internal/app/plan.go
@@ -86,24 +86,59 @@ func (p *plan) exec() {
 		log.Info("Nothing to execute")
 	}
 
+	wg := sync.WaitGroup{}
+	sem := make(chan struct{}, flags.parallel)
+	var fail bool
+	var priorities []int
+	pl := make(map[int][]orderedCommand)
 	for _, cmd := range p.Commands {
-		log.Notice(cmd.Command.Description)
-		result := cmd.Command.exec()
-		if cmd.targetRelease != nil && !flags.dryRun && !flags.destroy {
-			cmd.targetRelease.label()
+		pl[cmd.Priority] = append(pl[cmd.Priority], cmd)
+	}
+	for priority := range pl {
+		priorities = append(priorities, priority)
+	}
+	sort.Ints(priorities)
+
+	for _, priority := range priorities {
+		c := make(chan string, len(pl[priority]))
+		for _, cmd := range pl[priority] {
+			sem <- struct{}{}
+			wg.Add(1)
+			go func(cmd orderedCommand) {
+				defer func() {
+					wg.Done()
+					<-sem
+				}()
+				log.Notice(cmd.Command.Description)
+				result := cmd.Command.exec()
+				if cmd.targetRelease != nil && !flags.dryRun && !flags.destroy {
+					cmd.targetRelease.label()
+				}
+				if result.code != 0 {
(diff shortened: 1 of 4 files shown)


luisdavim:
By default it won't run any helm commands in parallel but that can be changed by setting the `-p` flag to a value bigger than 1 in which case it will run that amount of releases in parallel by priority.
I'd appreciate some help testing this in different setups from my own.
I think that in the future we should review the plan struct regarding how we store the commands and decisions and see if we can cleanup the code around adding them to the plan and executing the plan.

internal/app/plan.go
```diff
@@ -86,25 +86,45 @@ func (p *plan) exec() {
...
+				result := cmd.Command.exec()
+				if cmd.targetRelease != nil && !flags.dryRun && !flags.destroy {
+					cmd.targetRelease.label()
+				}
+				if result.code != 0 {
+					errorMsg := result.errors
+					if !flags.verbose {
+						errorMsg = strings.Split(result.errors, "---")[0]
+					}
+					log.Fatal(fmt.Sprintf("Command returned [ %d ] exit code and error message [ %s ]", result.code, strings.TrimSpace(errorMsg)))
```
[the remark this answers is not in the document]
luisdavim: yeah, I can do that. I'll update the PR so it only exits on failure at the end on each priority.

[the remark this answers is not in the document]
luisdavim: I've updated the PR, it will only stop at the end of each priority.

internal/app/plan.go
```diff
@@ -86,25 +86,57 @@ func (p *plan) exec() {
...
+	sem := make(chan struct{}, flags.parallel)
+	var fail bool
+	var priorities []int
+	pl := make(map[int][]orderedCommand)
 	for _, cmd := range p.Commands {
-		log.Notice(cmd.Command.Description)
-		result := cmd.Command.exec()
-		if cmd.targetRelease != nil && !flags.dryRun && !flags.destroy {
-			cmd.targetRelease.label()
+		priorities = append(priorities, cmd.Priority)
```
mkubaczyk, line 95: this can have more than one priority of the same value causing it to run more than once for specific charts. priorities array must contain unique values before it goes to loop in 99 line

internal/app/plan.go
```diff
@@ -86,25 +86,57 @@ func (p *plan) exec() {
...
+					wg.Done()
+					<-sem
+				}()
+				log.Notice(cmd.Command.Description)
+				result := cmd.Command.exec()
+				if cmd.targetRelease != nil && !flags.dryRun && !flags.destroy {
+					cmd.targetRelease.label()
+				}
+				if result.code != 0 {
+					errorMsg := result.errors
```
mkubaczyk, line 115: we should add chart name to the error message since for multiple errors this becomes unclear

internal/app/plan.go
luisdavim: good catch, I'll update the PR

luisdavim: done, the list of priorities is now deduplicated.

internal/app/plan.go
luisdavim: I've updated the message a few lines below.

luisdavim:
@mkubaczyk I've updated the PR, would you mind having another look?
Thanks

PR #442, opened 2020-04-27
Chart performance improvements

fmd:


internal/app/decision_maker.go
```diff
@@ -43,9 +43,10 @@ func buildState(s *state) *currentState {
 			}()
 			if flags.contextOverride == "" {
 				r.HelmsmanContext = getReleaseContext(r.Name, r.Namespace)
```
mkubaczyk, line 45: This will print the line for every app change which is unnecessary and will be too noisy

internal/app/helm_helpers.go
```diff
@@ -26,7 +29,13 @@ func helmCmd(args []string, desc string) command {
 
 // extractChartName extracts the Helm chart name from full chart name in the desired state.
 func extractChartName(releaseChart string) string {
-	cmd := helmCmd([]string{"show", "chart", releaseChart}, "Show chart information")
+	chart, ok := chartNameCache.Load(releaseChart)
+	if ok {
+		log.Info("Returning chart information for [ " + releaseChart + " ] from cache.")
```
mkubaczyk, line 34: This should be a verbose or debug imho, not info

internal/app/helm_helpers.go
```diff
@@ -53,6 +63,8 @@ func getHelmVersion() string {
 	if result.code != 0 {
 		log.Fatal("While checking helm version: " + result.errors)
 	}
+
+	log.Info("Helm version " + result.output + "")
```
mkubaczyk, line 67: Verbose imho

internal/app/release.go
```diff
@@ -219,9 +233,11 @@ func (r *release) validateChart(app string, s *state, c chan string) {
 // If chart is local, returns the given release version
 func (r *release) getChartVersion() (string, string) {
 	if isLocalChart(r.Chart) {
+		log.Info("Chart [ " + r.Chart + "] with version [ " + r.Version + " ] was found locally.")
```
mkubaczyk, line 236: Verbose

internal/app/main.go
```diff
@@ -9,7 +9,7 @@ const (
 	appVersion         = "v3.3.0"
 	tempFilesDir       = ".helmsman-tmp"
 	defaultContextName = "default"
-	resourcePool       = 10
+	resourcePool       = 100
```
mkubaczyk, line 12: But why? :-)

fmd:
one of the caches I wrote doesn't actually work

fmd:
i'm going to push a fix to this that makes the "cache" I put on releases actually work (it doesn't here), sorry i have to sanitise the changes as i go from an enterprise github fork and it's getting quite confusing

internal/app/main.go
fmd: there' something off about the concurrency that runs multiple helm commands, i don't think its speeding things up much, but i'll revert this for now

fmd:
closing to reopen

fmd:
@mkubaczyk https://github.com/Praqma/helmsman/pull/445

PR #445, opened 2020-04-29
Performance improvements

fmd:
In short, this PR makes `helmsman` call `helm` less. 

If we have 100 apps, and they are all on the same version of the same chart, helmsman would validate that chart 100 times in a row with identical calls to `helm`. This doesn't need to happen. I kept the idea of running helm commands concurrently a `resourcePool` number of times (more on `resourcePool` vs the `-p` flag later in this comment), but I have expanded it -- we concurrently ask helm about the charts (like we used to), but we only do so once per chart version pair (at least in `makePlan`). 

Compared to my previous PR, I am no longer trying to cache chart names and versions in an atomic `sync.Cache`. Since we went to the OS and called helm a bunch of times concurrently, they'd all check the cache and see the result wasn't there, then all the resources in the pool would attempt to fill the cache by calling the expensive `helm` command at the same time. It was still faster (because I decoupled `validateChart` from the release), but this is faster still.

I've updated the Dockerfile quite a bit too, I don't know if it'll pass CircleCI.

For our use case these changes seem to roughly double the speed of our deploys on helmsman. However, I haven't done any profiling (just a few very quick eyeball tests) and I haven't written any tests that test when there's lots of releases on a mixture of charts.

`resourcePool` now applies slightly differently to makePlan. I suggest we combine `resourcePool` with the `-p` flag so we can set the resourcePool number ourselves and have it apply consistently, but another time I want to look at how the concurrency on helm commands is performing across the board. Some helm commands still aren't run concurrently (like `r.checkChartDepUpdate()` which I haven't looked at yet), and i have no data on what kind of `resourcePool` number might be optimal, but this is a start.

This could definitely be DRYed up and refactored but I think these changes should provide some inroads into making helmsman more performant, especially when someone is deploying a lot of apps that share charts (as we are).

### DIFF
--- a/internal/app/decision_maker.go
+++ b/internal/app/decision_maker.go
@@ -44,8 +44,8 @@ func buildState(s *state) *currentState {
 			if flags.contextOverride == "" {
 				r.HelmsmanContext = getReleaseContext(r.Name, r.Namespace)
 			} else {
-				log.Info("Overriding Helmsman context for " + r.Name + " as " + flags.contextOverride)
 				r.HelmsmanContext = flags.contextOverride
+				log.Info("Overwrote Helmsman context for release [ " + r.Name + " ] to " + flags.contextOverride)
 			}
 			cs.releases[r.key()] = r
 		}(r)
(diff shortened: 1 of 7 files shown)


fmd:
Even with these changes, it's still a lot of calls to helm. There must be a way of doing this that minimises calls to helm further by design, but I'm quite busy so I can't think of it yet

fmd:
Screenshot 2020-04-30 at 12 24 30
For instance, we have a github check we wrote that can parse toml/yaml changes well enough to only search for packages that are being changed, so surely there's a way to not have to diff every release in a namespace?

Please let me know if there's an obvious solution that I haven't thought of, or if there's obvious reasons that this couldn't work, but it seems like we could make huge performance gains with something like a `helm multisearch` plugin that batches helm search commands into one json response, alongside something that passes in the releases that need to be diffed (we usually have a list of these by the time we call `helmsman`).

internal/app/decision_maker.go
```diff
@@ -60,18 +60,115 @@ func (cs *currentState) makePlan(s *state) *plan {
 
 	wg := sync.WaitGroup{}
 	sem := make(chan struct{}, resourcePool)
+	namesC := make(chan [2]string, len(s.Apps))
+	versionsC := make(chan [4]string, len(s.Apps))
+
+	// We store the results of the helm commands
+	extractedChartNames := make(map[string]string)
```
luisdavim, line 67: probably using a struct for the extracted chart names and versions would be cleaner but I guess we can do that in another iteation

internal/app/helm_helpers.go
```diff
@@ -53,6 +53,8 @@ func getHelmVersion() string {
 	if result.code != 0 {
 		log.Fatal("While checking helm version: " + result.errors)
 	}
+
+	log.Verbose("Helm version " + result.output)
```
mkubaczyk, line 57: there's no need to log that since the helm version is printed in cli.go:152

internal/app/release.go
```diff
@@ -172,79 +194,74 @@ func validateReleaseCharts(s *state) error {
...
 
 	result := cmd.exec()
 	if result.code != 0 {
-		return "", "Chart [ " + r.Chart + " ] with version [ " + r.Version + " ] is specified but not found in the helm repositories"
+		return "", "Chart [ " + chart + " ] with version [ " + version + " ] is specified but not found in the helm repositories"
 	}
 
-	chartVersions := make([]chartVersion, 0)
-	if err := json.Unmarshal([]byte(result.output), &chartVersions); err != nil {
+	cv := make([]chartVersion, 0)
```
mkubaczyk, line 247: out of curiosity - why are you changing more comprehensive variable names to simply not intuitive letters?

fmd:
@mkubaczyk i've seen your comments, i'll address them and clean this up in a day or so

mkubaczyk:
Please do. We are waiting with the release for your changes to join it as well. :-)

internal/app/release.go
fmd: fixed this

fmd:
@mkubaczyk fixed your comments.

I want to cache the downloaded tars that Docker `curl`s somehow, because at the moment it pushes new layers on every build which is wasting about the same amount of space on our hub as the old Dockerfile, but like you said it might not be too late to split out the Dockerfile changes into a new PR. Maybe we split the `helm-installer` part of the multistage build into its own Dockerfile that's ran with a different `make` command? I'd like to find a way to make it only have to curl the kubectl/helm/plugins `tar` once and then cache it for future builds. I don't want to keep pushing new layers with the same binary to the hub. What I've done here is a step towards that, removing the script was good I think, but we're not fully there.

internal/app/decision_maker.go
```diff
@@ -58,20 +58,124 @@ func buildState(s *state) *currentState {
 func (cs *currentState) makePlan(s *state) *plan {
 	p := createPlan()
 
+	var apps map[string]*release
+	if len(s.TargetMap) > 0 {
+		apps = s.TargetApps
+	} else {
+		apps = s.Apps
```
mkubaczyk, line 65: This results in not displaying apps not taken into account by -target flag, but we want them to still be displayed so the user knows what's whole scope of the DSF. Why did you add that?

fmd: my reasoning was that the display on apps we know don't need to be touched is costly -- is there a functional reason to add this in? I'll remove this from the PR shortly either way, just interested

mkubaczyk: the whole idea of DSF is to cover whole environment consisting of multiple apps. Even though you're having -target to one app, this is just a part of the whole env, so we want everyone to know it's still considered among all apps from this env, just marked as ignored, so there's no doubt helmsman sees the env fully

fmd: ok sure

fmd: @mkubaczyk fixed

internal/app/helm_helpers.go
```diff
@@ -25,8 +25,8 @@ func helmCmd(args []string, desc string) command {
 }
 
 // extractChartName extracts the Helm chart name from full chart name in the desired state.
-func extractChartName(releaseChart string, version string) string {
-	cmd := helmCmd([]string{"show", "chart", releaseChart, "--version", version}, "Show chart information for version "+version)
+func extractChartName(releaseChart string) string {
```
fmd, line 28: I reverted this change -- we no longer use this when we already have a version

fmd:
@mkubaczyk i've addressed your feedback and it looks like it passes the build again

mkubaczyk:
good, thanks! I'm merging it :-)

fmd:
@mkubaczyk a point of note: we should (I can, in a new PR) update the Dockerfile again and cache the initial container somehow, but only when we update the versions -- it would vastly improve the Docker layering to do so

mkubaczyk:
@fmd whatever you feel is right for the codebase, please propose it with MR. We love to see new things coming into the repo :-)

PR #484, opened 2020-06-02
Run helm tests also when upgrading chart

mkubaczyk:
To be honest I don't know why the helm tests were only considered for install action. :-)


mkubaczyk:
Yes. I did the basic confirmation and tests changes were not causing the deployment to happen (diff was ignoring it). :-)

PR #491, opened 2020-06-09
Fix missing valid parallel run for cmd exec after hooks were introduced

### LINKED ISSUE #490: Helmsman 3.4.x stops deployment leaveing other apps in pending state when one of apps fails
I think this bug was introduced with lifecycle hooks in https://github.com/Praqma/helmsman/commit/bb9531777ca48934d4c94871e2e3dea816c47c08 when parallel was abandoned.
If two apps with the same priority are being deployed in parallel, whenever one fails, second is left in pending state as the process exits immediately. There's no waiting for deployments to complete in order to validate the messages from the channel.

mkubaczyk:
Should fix #490 

This also fixes another issue when failing hooks for a chart were not failing its deployment and their results were just ignored

For the sake of readiness I let myself repeate some IFs in releaseWithHooks function

### DIFF
--- a/internal/app/plan.go
+++ b/internal/app/plan.go
@@ -1,6 +1,7 @@
 package app
 
 import (
+	"errors"
 	"fmt"
 	"net/url"
 	"sort"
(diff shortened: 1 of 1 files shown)


internal/app/plan.go
```diff
@@ -156,28 +128,58 @@ func (p *plan) exec() {
...
+			errors <- msg
+			log.Verbose(msg)
+			return
+		}
+	}
+}
+
 // execOne executes a single ordered command
-func execOne(cmd command, targetRelease *release) {
+func execOne(cmd command, targetRelease *release) (msg string, ok bool) {
```
luisdavim, line 161: why not just return an `error` instead of a `string` and a `bool`?

mkubaczyk: oh, I keep on confusing approaches between languages, sorry, of course I changed that

internal/app/plan.go
```diff
@@ -156,28 +129,62 @@ func (p *plan) exec() {
...
 	result := cmd.exec()
-
 	if result.code != 0 {
 		errorMsg := result.errors
 		if !flags.verbose {
 			errorMsg = strings.Split(result.errors, "---")[0]
 		}
 		if targetRelease != nil {
-			log.Fatal(fmt.Sprintf("Command for release [%s] returned [ %d ] exit code and error message [ %s ]", targetRelease.Name, result.code, strings.TrimSpace(errorMsg)))
+			return errors.New(
```
luisdavim, line 171: nit: this could be `fmt.Errorf`

mkubaczyk: did that, thanks

PR #505, opened 2020-08-11
helmRepos: remove "user:pass@" from url when adding repo

antoinedeschenes:
Since helmsman is calling `helm repo add` with `--username` and `--password`, this removes the duplicate username and password inline with the repository URL.


luisdavim (approved):
Thanks

Issue #414, opened 2020-03-01
Support for kustomize


sheerun:
Maybe it would be good idea to integrate with kustomize so it's possible to change used charts without forking them? https://github.com/kubernetes-sigs/kustomize 

The kustomization.yaml could be a part of per-app helmsman config

luisdavim:
and you'd suggest we'd install the helmcharts with `helm template` then run `kustomize` and finally `kubectl apply`? We'd be  loosing some of the features provided by helm or is there another way of using `kustomize` with helm?

sheerun:
You can use kustomize as `--post-renderer` without loosing any features https://helm.sh/docs/topics/advanced/

kferrone:
Oh wow thanks a ton @sheerun 
I'v been looking for an example like that. 
This would be a very useful feature.

luisdavim:
That's cool, I didn't knew helm supported that. Thanks.

luisdavim:
Can you try using the `helmFlags` option @kferrone mentioned and letting us know if it worked or if we need to make any changes to make it work?

kferrone:
@luisdavim  @sheerun 

Holy friggin smokes it worked like a charm!! 😂

`./helmsman.yaml`
```yaml
helmFlags: [
    "--post-renderer", 
    "./kustomize/kustomize.sh"
]
```

`./kustomize/kustomize.sh`
```bash
#!/bin/bash

cat <&0 > ./kustomize/all.yaml

kustomize build ./kustomize && rm ./kustomize/all.yaml
```

`./kustomize/kustomization.yaml`
```yaml
resources:
  - all.yaml
commonLabels:
  organization: praqma
```

To really get your kustomize game on set the following variable: 
`export PLUGIN_ROOT="./kustomize/plugins"`

Then make some kustomization plugins!

Happy Helming!!

sheerun:
helmFlags could would but I would really appreciate if it was possible to configure kustomize directly in per-app helmsman config (maybe also some global one)

kferrone:
I feel this feature request and #360 can be merged into one feature which simply utilizes kustomize. 
I delve deeper in a comment on #360 as to what I am imagining. 

@sheerun [Please read my comment](https://github.com/Praqma/helmsman/issues/360#issuecomment-597959093) and chime in with anything else you imagine.

luisdavim:
closed via https://github.com/Praqma/helmsman/pull/516

PR #516, opened 2020-09-24
post-renderer option

ndegory:
This PR implement the post renderer Helm feature as a first class citizen option in the Helmsman DSF app config.
It was already possible to use the post renderer via the HelmFlags array, as mentioned in #414 , this implementation adds:
- validation that the path of the executable is valid
- pass the post renderer option to helm diff

Signed-off-by: Nicolas Degory <ndegory@gmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,7 +64,7 @@ generate:
 .PHONY: generate
 
 repo:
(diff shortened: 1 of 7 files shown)


luisdavim (approved):
LGTM; Thanks

PR #521, opened 2020-10-02
Run helm tests as a last command of after commands from hooks

mkubaczyk:
I had various cases like:
`[2020-10-02T03:58:06.133Z] 2020-10-02 03:58:04 NOTICE: Running tests for release [ XXXX ] in namespace [ XXX ]
[2020-10-02T03:58:06.133Z] 2020-10-02 03:58:04 NOTICE: Upgrade release [ XXXX ] to version [ 1.2.0 ] in namespace [ XXX ]` causing helm tests to be run together with the release they are about to test as parallel flag was set to something > 1.
As tests are usually to verify what was deployed, I'd like to propose to run them as a last afterCommand reusing the scheme from hooks.

### DIFF
--- a/internal/app/release.go
+++ b/internal/app/release.go
@@ -292,21 +292,21 @@ func getChartVersion(chart, version string) (string, string) {
 }
 
 // testRelease creates a Helm command to test a particular release.
-func (r *release) test(p *plan) {
+func (r *release) test(afterCommands *[]command) {
 	cmd := helmCmd(r.getHelmArgsFor("test"), "Running tests for release [ "+r.Name+" ] in namespace [ "+r.Namespace+" ]")
-	p.addCommand(cmd, r.Priority, r, []command{}, []command{})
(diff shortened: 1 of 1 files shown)


internal/app/release.go
```diff
@@ -292,21 +292,21 @@ func getChartVersion(chart, version string) (string, string) {
 }
 
 // testRelease creates a Helm command to test a particular release.
-func (r *release) test(p *plan) {
+func (r *release) test(afterCommands *[]command) {
```
luisdavim, line 295: I don't think this needs to be a pointer here, slices are references to arrays.

mkubaczyk: Actually I did that on explicit as I'm changing the slice size causing it to allocate it as a new one. When tested, going with no pointers gives no results in the slice after function test() has been run, hence pointers

luisdavim (approved):
LGTM

PR #524, opened 2020-10-07
fix: helm test resets release labels

luisdavim:
After https://github.com/Praqma/helmsman/pull/521 the tests are actually executed and that seems to reset all the Helmsman labels in the helm release so I'm moving the labelling function to be executed after that.

Signed-off-by: Luis Davim <luis.davim@jet.com>

### DIFF
--- a/internal/app/plan.go
+++ b/internal/app/plan.go
@@ -133,6 +133,12 @@ func releaseWithHooks(cmd orderedCommand, wg *sync.WaitGroup, sem chan struct{},
 		wg.Done()
 		<-sem
 	}()
+	if cmd.targetRelease == nil && !flags.destroy {
+		err := fmt.Errorf("nil target release")
+		errors <- err
+		log.Verbose(err.Error())
+		return
+	}
 	for _, c := range cmd.beforeCommands {
 		if err := execOne(c, cmd.targetRelease); err != nil {
 			errors <- err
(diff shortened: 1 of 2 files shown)


internal/app/plan.go
```diff
@@ -133,6 +133,12 @@ func releaseWithHooks(cmd orderedCommand, wg *sync.WaitGroup, sem chan struct{},
 		wg.Done()
 		<-sem
 	}()
+	if cmd.targetRelease != nil && !flags.destroy {
+		err := fmt.Errorf("got an empty target release")
```
mkubaczyk, line 137: This says that it is an empty target release but the condition is actually checking if it’s not nil. Is this a mistake or I am just not getting this part? :-)

mkubaczyk (changes requested):

internal/app/plan.go
luisdavim: it's just the way I've worded it, I'll update it.

luisdavim: the idea here is to fail faster, allowing it to go to `execOne` would have it run the first command but then return with an error...

internal/app/plan.go
```diff
@@ -133,6 +133,12 @@ func releaseWithHooks(cmd orderedCommand, wg *sync.WaitGroup, sem chan struct{},
 		wg.Done()
 		<-sem
 	}()
+	if cmd.targetRelease != nil && !flags.destroy {
```
mkubaczyk, line 136: @luisdavim I mean this specific part. Don't you want it to be `if cmd.targetRelease == nil && !flags.destroy` if the error then says `fmt.Errorf("nil target release")`?

mkubaczyk (changes requested):

internal/app/plan.go
luisdavim: yes, this is checking if the target release is `nil` but only returns the error if the destroy flag is not set so we still try to run the uninstall command, keeping the previous behaviour for deletions.
I can remove the `&& !flags.destroy` part but the idea is to try to take the safest path...

luisdavim: sorry, I was blind and couldn't see the `!=` vs `==` :(

mkubaczyk (approved):
LGTM now :-)

PR #526, opened 2020-10-09
fix: handle default app name even when --skip-validation is used

antoinedeschenes:
A missing app.Name in the DSF would only default to the app state key when the state was validated. The `--skip-validation` flag would fallback to empty names, which unintendedly attempts to delete and install the release:

```
2020-10-09 09:13:23 NOTICE: Release [  ] version [ x.y.z ] will be installed in [ mynamespace ] namespace -- priority: -1000
2020-10-09 09:13:23 WARNING: Untracked release [ myapp ] found and it will be deleted -- priority: -1000
```

### DIFF
--- a/internal/app/cli.go
+++ b/internal/app/cli.go
@@ -240,6 +240,13 @@ func (c *cli) readState(s *state) {
 		}
 	}
 
+	// Default app.Name to state name when unset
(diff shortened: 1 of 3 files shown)


luisdavim (approved):
LGTM

Issue #512, opened 2020-09-17
 Error: failed to download "jupyterhub/binderhub" 


ctr26:
Hi all,

I'm trying to incorporate helmsman into my CI/CD for binderhub, unfortunately even a bare bones config fails "While getting chart information"

Looking at the debug I think the issue is that helmsman runs `helm show chart jupyterhub/binderhub` without a version number and for binderhub that returns nothing:

`helm show chart jupyterhub/binderhub --version 0.2.0-n219.hbc17443`

```
apiVersion: v1
dependencies:
- name: jupyterhub
  repository: https://jupyterhub.github.io/helm-chart
  version: 0.9.0
description: A helm chart to install Binder
name: binderhub
version: 0.2.0-n219.hbc17443
```

`helm show chart jupyterhub/binderhub`

`Error: failed to download "jupyterhub/binderhub" (hint: running `helm repo update` may help)`

Here's my config:

```
namespaces:
  binderhub-staging:

helmRepos:
  jupyterhub: "https://jupyterhub.github.io/helm-chart/"

apps:
  binderhub-staging:
    valuesFiles:
      - "binderhub/config.yaml"
      - "binderhub/secret.yaml"
    name: "binderhub-staging"
    namespace: "binderhub-staging"
    chart: "jupyterhub/binderhub"
    version: "0.2.0-n219.hbc17443"
    priority: -3

```

mkubaczyk:
If the version of a chart is not being passed to helm show chart command, it seeks for latest one. Somehow your helm repo does not let helm know what the latest is. How is your repo deployed? I checked that on private repo of mine and there's everything fine event though I didn't have to explicitly mark anything as latest, it just takes the latest one date-wise I suppose.

ctr26:
Binderhub (not my chart) doesnt install unless you specify a version latest. I'm assuming that's because it's helm3 specific and helm3 requires a version number.BestCraigOn 18 Sep 2020 08:11, Mateusz Kubaczyk <notifications@github.com> wrote:
If the version of a chart is not being passed to helm show chart command, it seeks for latest one. Somehow your helm repo does not let helm know what the latest is. How is your repo deployed? I checked that on private repo of mine and there's everything fine event though I didn't have to explicitly mark anything as latest, it just takes the latest one date-wise I suppose.

ctr26:
[...]

PR #528, opened 2020-10-21
Fix for helm charts without a latest version? 

ctr26:
https://github.com/Praqma/helmsman/issues/512

I've had a crack at fixing the bug I described here for the case where helm charts don't have a latest version.


antoinedeschenes:
You should probably use the output from "getChartVersion" when calling "extractChartName", version can be a semver expression otherwise

antoinedeschenes:
This also allows helmsman to work when a repository has a broken latest release

mkubaczyk:
closing in favour of #532 :-)

ctr26:
https://github.com/Praqma/helmsman/pull/532#issuecomment-723256091

This pull request may need reopening.

Issue #487, opened 2020-06-05
Helmsman skipping deploys where same chart version is already deployed


kbirger:
> DECISION: release [ my-app ] is desired to be enabled and is currently enabled. Nothing to do here! -- priority: 0

That is the message I get. I am not sure if there is a way to get this to work with existing settings, or if it would require a feature update.

There are cases where we want to redeploy the same version of a chart because:

1. We had to manipulate the chart's k8s resources directly on the cluster for troubleshooting and want to roll back the manual changes (my actual scenario)
2. We want to re-deploy the same chart with different settings (have not actually tested this yet, but suspect it would behave the same as I **did** test by DELETING everything in k8s but leaving the chart released).

It is not desirable to have a check under these circumstances which simply short circuits the deploy because the versions match up.

So is there a way to configure this with the existing version?

Using the latest Helmsman and latest Helm v2 version (We are not on v3 yet)

mkubaczyk:
helm-diff is deciding on it based on the differences between current state on the cluster and the desired state from interpolated templates from your chart. If there's no change, it won't do anything, so that is expected. You can redeploy it by disabling it and then enabling again or by passing some value that is different comparing to the one from previous deployment, so the diff will discover that change.

kbirger:
Sure, I could do all of those things and more, but isn't the purpose of this tool to be able to declaratively set an environment into a state which is declaratively defined? 

If I have to manually manipulate settings on my state just to get it to do what I want, it sort of defeats the purpose. I believe that is the reason that Helm provides a --force flag, so that you can make your operation idempotent regardless of the current state of things. In fact, I see that this flag is supported by Helmsman. Unfortunately in my case, Helm is not even being executed since Helmsman is optimistically assuming that the state is correct.

My goal is to have this run unattended, so I can't be marking settings just to fool helm-diff into seeing the truth: that my environment is **not** in the state that was last recorded.

mkubaczyk:
It's a matter of how Helm is able to discover changes you did manually rather Helmsman itself. I'd say it's undesired to manually mess with anything that is being managed by Helm, at least with current state of Helm where it has hard times discovering such changes happening.

kbirger:
Yes, I understand the conundrum, but is it not also interfering with Helm by preempting it in this instance? I'm not saying that the existing behavior is undesired in every case, but do you not think it's a valid usecase for Helmsman to use it on a non-prod environment that may have manual changes due to the need to troubleshoot?

[...]

PR #534, opened 2020-10-27
Add --always-upgrade flag

antoinedeschenes:
Add flag to run an upgrade even if helm diff finds no changes, allowing to overwrite manual changes made to the cluster (works around the missing 3-way diff in the helm-diff plugin as described in #487 )

### DIFF
--- a/internal/app/decision_maker.go
+++ b/internal/app/decision_maker.go
@@ -406,7 +406,10 @@ func (cs *currentState) inspectUpgradeScenario(r *release, p *plan, chartName, c
 				" ]. Delete of the current release will be planned and new chart will be installed in namespace [ "+
 				r.Namespace+" ]", r.Priority, change)
 		} else {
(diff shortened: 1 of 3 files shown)


internal/app/decision_maker.go
```diff
@@ -409,6 +409,9 @@ func (cs *currentState) inspectUpgradeScenario(r *release, p *plan, chartName, c
 			if diff := r.diff(); diff != "" {
 				r.upgrade(p)
 				p.addDecision("Release [ "+r.Name+" ] will be updated", r.Priority, change)
+			} else if flags.alwaysUpgrade {
```
luisdavim, line 412: why not do this before the diff and avoid the unnecessary operation?

antoinedeschenes: good point.

mkubaczyk (approved):
LGTM now

Issue #539, opened 2020-11-05
Helm/kubectl command fails/times out on checking cluster status on GKE


mkubaczyk:
This is something that usually happens on GKE, but there were also confirmed cases for EKS and AKS as well.
These are the examples of it:
```
2020-11-03 10:42:57 CRITICAL: Command returned with exit code: 1. And error message: Error: Failed to get release <cut> in namespace <cut>: exit status 1: Error: Kubernetes cluster unreachable: Get "https://<cut>/version?timeout=32s": net/http: request canceled (Client.Timeout exceeded while awaiting headers)

2020-11-05 07:31:33 CRITICAL: Failed to list all releases: Error: list: failed to list: stream error when reading response body, may be caused by closed connection. Please retry. Original error: stream error: stream ID 3; INTERNAL_ERROR

2020-11-04 13:05:43 CRITICAL: error: stream error when reading response body, may be caused by closed connection. Please retry. Original error: stream error: stream ID 1; INTERNAL_ERROR

2020-11-04 13:19:49 CRITICAL: Unable to connect to the server: read tcp <cut>:49302-><cut>:443: read: connection timed out
```
As on GKE/EKS/AKS you usually can not directly edit control plane and its timeouts, I'd see a retry option for specific commands done by Helmsman to mitigate the issue, that is: listing releases in namespace and getting secrets with helm release status.

PR #540, opened 2020-11-05
Add retryExec func for getting cluster status methods

mkubaczyk:
I hope this will mitigate the problems from #539

I've added retryExec to run `Getting Helmsman-managed releases` and `Listing all existing releases` with retries. It won't affect anyone not experiencing the same issues, but for those that do (like https://github.com/helm/helm/issues/7997) it should at least do its best. 

In the next iteration we can implement exponential backoff instead of waiting fixed time as it is now, but first I'd like to release this change and collect some examples.

### DIFF
--- a/internal/app/command.go
+++ b/internal/app/command.go
@@ -3,6 +3,7 @@ package app
 import (
 	"bytes"
 	"fmt"
+	"math"
 	"os/exec"
 	"strings"
 	"syscall"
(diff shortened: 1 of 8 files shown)


internal/app/command.go
```diff
@@ -25,6 +27,27 @@ func (c *command) String() string {
 	return c.Cmd + " " + strings.Join(c.Args, " ")
 }
 
+// runs exec command with retry
+func (c *command) retryExec(attempts int) exitStatus {
+	var result exitStatus
+	var errMsg string
+	err := retry(attempts, 2*time.Second, func() (err error) {
```
luisdavim, line 34: I don't think we need this many levels of wrapping. We could just call the exec function here in a loop and parse the exit code directly. If the goal is to have something more generic we can simply use something like https://github.com/avast/retry-go

mkubaczyk: ok, I went easier with the loop inside of retryExec
and also added math.Pow to time sleep there

luisdavim (approved):
LGTM

internal/app/command.go
```diff
@@ -25,6 +28,29 @@ func (c *command) String() string {
 	return c.Cmd + " " + strings.Join(c.Args, " ")
 }
 
+// runs exec command with retry
+func (c *command) retryExec(attempts int) exitStatus {
+	var result exitStatus
+
+	for i := 0; ; i++ {
```
luisdavim, line 35: why not:
```go
for i := 0; i >= (attempts - 1); i++ {
// ...
}
```

internal/app/command.go
```diff
@@ -25,6 +28,29 @@ func (c *command) String() string {
...
+		result = c.exec()
+		if result.code == 0 {
+			return result
+		}
+		if i >= (attempts - 1) {
```
luisdavim, line 40: we don't need this here if we set the condition in the loop definition....

internal/app/command.go
mkubaczyk: I kept things this way so I can use break before I print the Retrying message. Going with your suggestion I still need to add condition inside of the loop to check if that is going to be the last attempt so I don't print that message in the end

mkubaczyk: Let me know if you're fine with what I've changed now @luisdavim

luisdavim (approved):
LGTM

PR #542, opened 2020-11-09
Add retryExec to diff()

mkubaczyk:
I forgot to wrap helm diff command with retryExec.
I'd also like to release it as 3.5.1 if possible.


luisdavim (approved):
👍

Issue #530, opened 2020-10-24
extractChartName fails when only devel versions of a chart are published


ndegory:
Helmsman fails to get the chart information when only devel version (such x.y.z-rc0) are published.

```
2020-10-22 19:56:51 INFO: Adding helm repository [ my-repo ]
2020-10-22 19:56:54 INFO: Validating [ my-repo/my-chart ] chart's version [ 0.22.0-SNAPSHOT.15806 ] availability
6932020-10-22 19:56:54 DEBUG: helm search repo my-repo/my-chart --version 0.22.0-SNAPSHOT.15806 -l
2020-10-22 19:57:25 INFO: Extracting chart information for [ my-repo/my-chart ]
8172020-10-22 19:57:25 DEBUG: helm show chart my-repo/my-chart
8182020-10-22 19:57:26 INFO: Getting latest non-local chart's version my-repo/my-chart-0.22.0-SNAPSHOT.15806
8192020-10-22 19:57:26 DEBUG: helm search repo my-repo/my-chart --version 0.22.0-SNAPSHOT.15806 -o json
2020-10-22 19:57:28 CRITICAL: While getting chart information: Error: failed to download "my-repo/my-chart" (hint: running `helm repo update` may help)
```

This doesn't happen when a X.Y.Z version of the chart is available on the repo.

It seems to come from the `extractChartName(releaseChart string)` function which use the `helm show chart` command without the --devel argument.

PR #545, opened 2020-11-17
refactor chart validation and plan chart info extraction

antoinedeschenes:
This fetches both chart version and information simultaneously from `helm show chart {chart} --version {v}`. Chart information is extracted ahead of plan and used for chart validation.

- Works much faster than `helm search repo` and supports local charts
- Uses version when fetching chart name (should still work with #530)
- Still works when latest repository release is broken

### DIFF
--- a/internal/app/decision_maker.go
+++ b/internal/app/decision_maker.go
@@ -1,7 +1,6 @@
 package app
 
 import (
-	"fmt"
 	"regexp"
 	"strings"
 	"sync"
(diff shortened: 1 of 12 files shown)


antoinedeschenes:
@luisdavim I think I still have some cleanup/more work to do, but this makes sense overall? 

I'm really under the impression the `validateChart` is duplicate code, it's doing the exact same thing in a less efficient way (finding a matching version for the chart)

luisdavim:
> @luisdavim I still have some cleanup to do, but this makes sense overall? 
> 
> I'm really under the impression the code from `validateChart` duplicates this, as it's doing the exact same thing (throw an error if the chart version can't be found)

I haven't checked the code yet but you're probably right.

internal/app/decision_maker.go
```diff
@@ -124,39 +116,30 @@ func (cs *currentState) makePlan(s *state) *plan {
...
 				// ideally you just process all helm commands "in a background pool", they return channels,
 				// and we would be looping over select statements until we had the results of the helm commands we wanted.
-				n, e := getChartVersion(chart, version)
-				m := ""
-				if e != nil {
-					m = e.Error()
-				}
-				versionsC <- [4]string{chart, version, n, m}
+				i, e := getChartInfo(chart, version)
+				infoC <- struct {
```
luisdavim, line 120: maybe instead of a channel we can just use a `sync.Mutex` and update the `extractedChartInfo` map here directly.

antoinedeschenes: done

antoinedeschenes:
Current and previous behavior is to remove the release from the plan and print `ERROR: Chart [ stable/chartmus ] with version [ 2.12.0 ] is specified but not found in the helm repositories`, but keeps going.

What about extracting chart information one time outside of makePlan in the global state, and crash unless skipValidation is enabled? (removing validateReleaseCharts)

luisdavim:
> Current and previous behavior is to remove the release from the plan and print `ERROR: Chart [ stable/chartmus ] with version [ 2.12.0 ] is specified but not found in the helm repositories`, but keeps going.
> 
> What about extracting chart information one time outside of makePlan in the global state, and crash unless skipValidation is enabled? (removing validateReleaseCharts)

yes, that could work, if we need to extract that info anyway we might as well do it early and use it for the validation...

antoinedeschenes:
excellent, it works so far, but I'll need to add a version test as `helm show chart` seems to ignore the version flag on local charts

antoinedeschenes:
Tested so far:

antoinedeschenes:
Not sure what is failing the tests on CircleCI (no test cases are actually broken)

Manually tested cases:
- Repo from issue #512 still works (dev version)
- Doesn't fetch charts for non-targeted releases
- Skip-validation still skips releases with chart errors
- Exits on chart error without skip-validation flag
- Wildcard or exact version in DSF returns latest matching version for repository, error on mismatch
- Wildcard or exact version in DSF returns local chart if version matches, error on mismatch
- Chart info can still be extracted when latest release is broken (ex. latest release is in repo index but missing from repository)
- Doesn't pass validation with a partially matching chart name (unlike `helm search repo`)

luisdavim:
today there was some partial outage on circleCI and the googleapis storage and the setup of for the tests was intermittently failing....

internal/app/helm_helpers.go
```diff
@@ -4,21 +4,28 @@ import (
...
 	"net/url"
 	"path/filepath"
-	"regexp"
 	"strings"
 
 	"github.com/hashicorp/go-version"
 
 	"github.com/Praqma/helmsman/internal/gcs"
+
+	"github.com/Masterminds/semver/v3"
```
luisdavim, line 16: in other places we're using `github.com/hashicorp/go-version` can you use that here as well? or change the other places to use this one....

luisdavim:
> today there was some partial outage on circleCI and the googleapis storage and the setup of for the tests was intermittently failing....

Looks like the tests are actually failing, this is the relevant part from the logs:

```
2020-11-18 22:08:14 INFO: validating environment variables in ../../examples/example.toml
2020-11-18 22:08:14 CRITICAL: stat .helmsman-tmp: no such file or directory
```

internal/app/helm_helpers.go
antoinedeschenes: Looks like it has the exact same methods, I'll switch to go-version

antoinedeschenes:
> > today there was some partial outage on circleCI and the googleapis storage and the setup of for the tests was intermittently failing....
> 
> Looks like the tests are actually failing, this is the relevant part from the logs:
> 
> ```
> 2020-11-18 22:08:14 INFO: validating environment variables in ../../examples/example.toml
> 2020-11-18 22:08:14 CRITICAL: stat .helmsman-tmp: no such file or directory
> ```

Ok `make test` works locally, I'll try to see what is going on in the docker tests

antoinedeschenes:
Ok got it, the `state.expand()` method was called in the state_file_tests without a proper setup. It requires the presence of `.helmsman-tmp/`. It worked only because a previous test had it's teardown commented out. 

This time CircleCI crashed during `apk add`, someone will have to restart the job.

luisdavim (approved):
LGTM; Thanks 👍

Issue #502, opened 2020-07-08
Changing chart results in invalid `helm uninstall` command


sarumont:
Helmsman `3.4.2`

I am updating a helm release in our Helmsman-managed environment to pull a new chart. Helmsman detects the change properly, but when it issues the `uninstall` command, it is missing the namespace:

```
2020-07-08 02:47:28 NOTICE: -------- PLAN starts here --------------
...
2020-07-08 02:47:28 NOTICE: Release [ gateway ] is desired to use a new chart [ neverfail/dana ]. Delete of the current release will be planned and new chart will be installed in namespace [ default ] -- priority: 0
...
2020-07-08 02:47:28 NOTICE: -------- PLAN ends here --------------
2020-07-08 02:47:28 INFO: Printing the commands of the current plan ...
helm uninstall --namespace  gateway
helm upgrade gateway neverfail/dana --install --version 0.2.2 --namespace default --set fullnameOverride=gateway --set global.vault.name=nf-centralus-***-default --set global.vault.prefix=*** --set devMode=false --set image.tag=1.2.1 --set schema.image.tag=1.0.0 
2020-07-08 02:47:28 INFO: Executing plan... 
2020-07-08 02:47:28 NOTICE: Delete release [ gateway ] in namespace [  ]
2020-07-08 02:47:28 DEBUG: helm uninstall --namespace  gateway
2020-07-08 02:47:28 CRITICAL: Command for release [gateway] returned [ 1 ] exit code and error message [ Error: "helm uninstall" requires at least 1 argument
```

Note: `helm uninstall --namespace  gateway` doesn't have the `default` namespace specified, so `--namespace` eats the name of the deployment, leaving `helm` one argument short. :)

bergerx:
Please also consider #538 when working on this.

erpel:
I'm also seeing this on v3.4.6 even with a namespace defined in the desired state.
The --namespace flag is immediately followed by the release name.

bergerx:
Ah just realised this is exactly same issue with https://github.com/Praqma/helmsman/issues/538. Please review that one and dont just fix the uninstall. Now that helm supports upgrading an existing release with a different chart, Helmsman doesn't need to (and should not) uninstall the old release at all.

erpel:
If one were to add special handling for this case, it might be a good idea to supply/set "--set nameOverride=" automatically, though I do not know enough about all the use cases to know if this is actually appropriate as default behaviour.

luisdavim:
[...]

Issue #538, opened 2020-11-04
Helmsman seems to be unintentionally running as it should be on chart renames


bergerx:
Now that the helm organisation is deprecating the stable and incubator chart repos, many charts have been moving around and some are being renamed as part of the effort as well (e.g. `stable/prometheus-operator` --> `prometheus-community/kube-prometheus-stack`).

So being able to upgrade charts with `nameOverrides` is important to in-place upgrade the releases without losing the pointers to persistent data. Helm supports this scenario as of Helm 3 I guess. Apparently Helmsman still has some kind of special Helm 2 logic to support the chart rename scenario, in which case Helmsman is trying to uninstall and install which is not needed at all now.

**Luckily Helmsman fails to delete the release with the updated chart name and ends up doing the exact same thing with Helm 3 as an unintended result.**

Helmsman reports misleading "Delete release" and "helm uninstall" failure.

I'm using helmsman version: v3.4.6.

Now lets simulate a chart rename scenario:
```bash
helm create testing-chart

# a renamed version of the chart
cp -ar testing-chart testing-chart-renamed
find testing-chart-renamed -type f | xargs  sed -i 's/testing-chart/testing-chart-renamed/g'
```

What helm does:
```bash
helm upgrade --install \
  testing-helm \
  testing-chart  # revision 1
# Release "testing-helm" does not exist. Installing it now.
# REVISION: 1
kubectl get deployment  # testing-helm-testing-chart

# upgrade the existing release with the renamed chart
helm upgrade --install \
  --set nameOverride=testing-chart \
  testing-helm \
  testing-chart-renamed   # same helm release, just revision 2
kubectl get deployment  # testing-helm-testing-chart, still the same deployment, spec updated 
```

What helmsman does in the same scenario:
```bash 
cat <<EOF > helmsman.yaml
namespaces:
  default:
apps:
  testing-helmsman:
    chart: testing-chart
    version: 0.1.0
    namespace: default
    enabled: true
EOF
helmsman --apply -f helmsman.yaml
# 2020-11-04 21:52:54 NOTICE: -------- PLAN starts here --------------
# 2020-11-04 21:52:54 NOTICE: Release [ testing-helmsman ] version [ 0.1.0 ] will be installed in [ default ] namespace -- priority: 0
# 2020-11-04 21:52:54 NOTICE: -------- PLAN ends here --------------
# 2020-11-04 21:52:54 INFO: Executing plan... 
[...]

PR #546, opened 2020-11-18
fix: no special logic for rename is needed using helm v3

luisdavim:
This should fix: #502 and #538

### DIFF
--- a/internal/app/decision_maker.go
+++ b/internal/app/decision_maker.go
@@ -296,32 +296,8 @@ func (cs *currentState) inspectUpgradeScenario(r *release, p *plan, c *chartInfo
 		return
 	}
 
-	if r.Namespace == rs.Namespace {
-		r.Version = c.Version
-
-		if c.Name == rs.getChartName() && r.Version != rs.getChartVersion() {
-			// upgrade
-			r.diff()
-			r.upgrade(p)
-			p.addDecision("Release [ "+r.Name+" ] will be upgraded", r.Priority, change)
-
-		} else if c.Name != rs.getChartName() {
-			r.reInstall(p)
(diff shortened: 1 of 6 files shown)


internal/app/decision_maker.go
```diff
@@ -296,38 +296,40 @@ func (cs *currentState) inspectUpgradeScenario(r *release, p *plan, c *chartInfo
...
+		return
+	}
+
+	if flags.alwaysUpgrade || r.Version != rs.getChartVersion() || c.Name != rs.getChartName() {
+		// Version changed
+		r.upgrade(p)
+		p.addDecision("Release [ "+r.Name+" ] will be upgraded", r.Priority, change)
+		return
+	}
+	if diff := r.diff(); diff == "" && !flags.alwaysUpgrade {
```
antoinedeschenes, line 329: Could remove `&& !flags.alwaysUpgrade` here. 

I'd switch the `if` condition to `diff != ""`  so that the noop decision happens when no cases match

luisdavim: Ok, I've actually went back and forth with both versions and couldn't decide which would be the best option. the `&& !flags.alwaysUpgrade` was there just for safety....

internal/app/helm_helpers.go
```diff
@@ -55,10 +55,13 @@ func getChartInfo(chartName, chartVersion string) (*chartInfo, error) {
 	}
 
 	constraint, _ := version.NewConstraint(chartVersion)
```
antoinedeschenes, line 57: good catch, forgot to handle these, looks like we missed the constraint error as well

antoinedeschenes:
LGTM, makes sense to switch to `Mastermind/semver` if `hashicorp/go-version` doesn't support the full semver spec, we can expect consistent version checking with helm now (uses Masterminds/semver/v3)

PR #547, opened 2020-11-19
-update-deps: run "helm dep up" concurrently

antoinedeschenes:
also add '--skip-refresh' flag as 'helm repo update' is already run by helmsman, and is otherwise not safe to use in parallel


luisdavim (approved):
OK

PR #548, opened 2020-11-19
fetch "helm plugins list" once

antoinedeschenes:
Stops helmsman from executing `helm plugin list` for every targets. List is cached on first call.
Also add a retry on curl as the CI build seems to fail often there.

### DIFF
--- a/internal/app/helm_helpers.go
+++ b/internal/app/helm_helpers.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net/url"
 	"path/filepath"
+	"regexp"
 	"strings"
 
 	"gopkg.in/yaml.v2"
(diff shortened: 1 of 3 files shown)


internal/app/helm_helpers.go
```diff
@@ -93,15 +94,25 @@ func checkHelmVersion(constraint string) bool {
 // helmPluginExists returns true if the plugin is present in the environment and false otherwise.
 // It takes as input the plugin's name to check if it is recognizable or not. e.g. diff
 func helmPluginExists(plugin string) bool {
-	cmd := helmCmd([]string{"plugin", "list"}, "Validating that [ "+plugin+" ] is installed")
-
-	result := cmd.Exec()
-
-	if result.code != 0 {
-		return false
+	if plugins == nil {
```
luisdavim, line 97: This will only ever run once so if we check for more than one plugin, only the first one will be checked for and the remaining ones will  always return false.

internal/app/main.go
```diff
@@ -18,6 +18,7 @@ var (
 	settings   *config
 	curContext string
 	log        = &Logger{}
+	plugins    map[string]bool
```
luisdavim, line 21: not a bit fan of global variables, I've been trying to reduce the usage of global vars in this code...

luisdavim (commented):
Do we really need this?
The `diff` plugin is only checked once, in `cli.parse()` we could add some logic there to identify if the other plugins are needed and check them there instead of where they are currently being checked.

internal/app/helm_helpers.go
antoinedeschenes: It's filling the global map with all results from `helm plugin list` except the "NAME" header

antoinedeschenes:
> Do we really need this?
> The `diff` plugin is only checked once, in `cli.parse()` we could add some logic there to identify if the other plugins are needed and check them there instead of where they are currently being checked.

We're checking for secrets and gcs every time they're specified in an app's state. Actually we're also checking for hiera-eyaml in loops when EyamlEnabled is set (via ToolExists). Maybe there should be some place (in cli, state.Settings or else) checking and caching "capabilities" (eyaml, gcs, diff, PostRenderer)

internal/app/main.go
antoinedeschenes: It doesn't look like it belongs to any of them, but seeing how state.StorageBackend is used, I suppose it could be in the state. Otherwise it could use it's own package with a global var, but that seems overkill for a small use case

luisdavim:
> > Do we really need this?
> > The `diff` plugin is only checked once, in `cli.parse()` we could add some logic there to identify if the other plugins are needed and check them there instead of where they are currently being checked.
> 
> We're checking for secrets and gcs every time they're specified in an app's state. Actually we're also checking for hiera-eyaml in loops when EyamlEnabled is set (via ToolExists). Maybe there should be some place (in cli, state.Settings or else) checking and caching "capabilities" (eyaml, gcs, diff, PostRenderer)

yes, that's what I meant we can check the flags and the sate to detect if those plugins will be needed and check for them early in `cli.parse()` or maybe `cli.readState()` and stop checking every time they are used, like we do with the `diff` plugin.

internal/app/main.go
luisdavim: I think we can re-engineer this so we don't need to call this method for the same plugin multiple times.

internal/app/helm_helpers.go
luisdavim: yes, you're right, I guess the `plugin` threw me off.

antoinedeschenes:
Will reopen once ready

PR #549, opened 2020-11-24
fix: isValidFile error for hooks should not lead to a file resolve.

jmontroy90:
A first cut of shell hooks was released in https://github.com/Praqma/helmsman/pull/546 -- looks like there's a bug in how `isValidFile` is being used for hooks. I think it should be a `==` instead of `!=` in a few of these places.

**Repro**

DSF:
```
<snip>
apps:
    sample:
        name: sample
        valuesFiles:
          - sample-values.yaml
        <snip>
        hooks:
            postInstall: my-legit-command-here
```

Error:
```
...
2020-11-23 22:18:07 INFO: sample-values.yaml will be used from local file system.
2020-11-23 22:18:07 INFO: cluster-meta.yaml will be used from local file system.
2020-11-23 22:18:07 CRITICAL: open : no such file or directory
```

This is definitely a hotfix, though. Leaving as WIP while I continue to validate and open to suggestions of more durable solutions here.

cc @luisdavim

### DIFF
--- a/internal/app/release.go
+++ b/internal/app/release.go
@@ -485,13 +485,13 @@ func (r *release) getHookCommands(hookType, ns string) []hookCmd {
 	var cmds []hookCmd
 	if _, ok := r.Hooks[hookType]; ok {
 		hook := r.Hooks[hookType].(string)
-		if err := isValidFile(hook, []string{".yaml", ".yml", ".json"}); err != nil {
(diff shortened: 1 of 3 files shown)


luisdavim:
you're right, I think think I did some late refactoring in the PR where I introduced the new hook types and changed the way `isValidFile` works but forgot about changing the rest of the code :(
I think I'll merge this and release a hot fix and then we can think about improving this and adding some tests

luisdavim (approved):

PR #551, opened 2020-11-24
Initialize namespaces defined without any properties on init

### LINKED ISSUE #550: Panic for operation on namespace defined as an empty map in 3.6.1
If namespaces in DSF are defined like that:
`namespaces:
   mynamespace:
   mynamespace2:
`

then running the 3.6.0 fails with:

`panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x38 pc=0x18f3d7c]

goroutine 1 [running]:
github.com/Praqma/helmsman/internal/app.(*namespace).Disable(...)
	github.com/Praqma/helmsman/internal/app/namespace.go:50
github.com/Praqma/helmsman/internal/app.(*state).disableUntargetedApps(0xc0004d0000, 0x0, 0x0, 0x0, 0xc00023dcb0, 0x1, 0x1)
	github.com/Praqma/helmsman/internal/app/state.go:305 +0x49c
github.com/Praqma/helmsman/internal/app.(*cli).readState(0x2223900, 0xc0004d0000)
	github.com/Praqma/helmsman/internal/app/cli.go:248 +0x5aa
	github.com/Praqma/helmsman/internal/app/main.go:38 +0xc5
main.main()
	command-line-arguments/main.go:8 +0x25`

while trying to access .disable param of it

mkubaczyk:
Fixes #550 
It should go as 3.6.1 asap imho

### DIFF
--- a/internal/app/state.go
+++ b/internal/app/state.go
@@ -68,6 +68,15 @@ func (s *state) setDefaults() {
 	}
 }
 
+func (s *state) initializeNamespaces() {
+	for nsName, ns := range s.Namespaces {
+		if ns == nil {
+			ns = &namespace{}
+			s.Namespaces[nsName] = ns
+		}
+	}
+}
+
 // validate validates that the values specified in the desired state are valid according to the desired state spec.
 // check https://github.com/Praqma/helmsman/blob/master/docs/desired_state_specification.md for the detailed specification
 func (s *state) validate() error {
--- a/internal/app/state.go
+++ b/internal/app/state.go
@@ -71,8 +71,7 @@ func (s *state) setDefaults() {
 func (s *state) initializeNamespaces() {
 	for nsName, ns := range s.Namespaces {
 		if ns == nil {
-			ns = &namespace{}
-			s.Namespaces[nsName] = ns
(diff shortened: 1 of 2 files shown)


internal/app/state.go
```diff
@@ -68,6 +68,15 @@ func (s *state) setDefaults() {
 	}
 }
 
+func (s *state) initializeNamespaces() {
+	for nsName, ns := range s.Namespaces {
+		if ns == nil {
+			ns = &namespace{}
```
luisdavim, line 74: no need for the extra var, why not just `s.Namespaces[nsName] = &namespace{}`?

luisdavim (commented):
We should also check if `s.Namespaces == nil` and create the map if that's the case.

internal/app/state.go
mkubaczyk: right

mkubaczyk:
empty s.Namespaces is correctly recognised with code according to my local run for empty namespaces setting

luisdavim (approved):

PR #552, opened 2020-11-24
test: add test cases for readState

luisdavim:
More tests can be added but this is a starting point

### DIFF
--- a/internal/app/cli.go
+++ b/internal/app/cli.go
@@ -164,7 +164,7 @@ func (c *cli) parse() {
 		os.Setenv("KUBECONFIG", c.kubeconfig)
 	}
 
-	if !ToolExists("kubectl") {
+	if !ToolExists(kubectlBin) {
 		log.Fatal("kubectl is not installed/configured correctly. Aborting!")
 	}
 
@@ -191,13 +191,13 @@ func (c *cli) parse() {
 }
 
 // readState gets the desired state from files
-func (c *cli) readState(s *state) {
+func (c *cli) readState(s *state) error {
 	// read the env file
 	if len(c.envFiles) == 0 {
 		if _, err := os.Stat(".env"); err == nil {
 			err = godotenv.Load()
 			if err != nil {
-				log.Fatal("Error loading .env file")
+				return fmt.Errorf("error loading .env file: %w", err)
 			}
 		}
 	}
@@ -205,7 +205,7 @@ func (c *cli) readState(s *state) {
 	for _, e := range c.envFiles {
 		err := godotenv.Load(e)
 		if err != nil {
-			log.Fatal("Error loading " + e + " env file")
+			return fmt.Errorf("error loading %s env file: %w", e, err)
 		}
 	}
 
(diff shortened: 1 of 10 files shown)


internal/app/main.go
```diff
@@ -35,7 +35,20 @@ func Main() {
 		defer s.cleanup()
 	}
 
-	flags.readState(&s)
+	if err := flags.readState(&s); err != nil {
+		log.Fatal(err.Error())
+	}
+
+	if len(flags.target) > 0 && len(s.TargetMap) == 0 {
+		log.Info("No apps defined with -target flag were found, exiting")
```
jmontroy90, line 43: Should this be logged at `Error` or something?

internal/app/command_test.go
```diff
@@ -5,6 +5,44 @@ import (
...
+		want bool
+	}{
+		{
+			name: "test case 1 -- checking helm exists.",
+			args: args{
+				tool: helmBin,
+			},
+			want: true,
+		}, {
+			name: "test case 2 -- checking kubectl exists.",
```
jmontroy90, line 24: It's weird to mandate `kubectl` and `helm` are installed...this means these tests can't predictably be run locally, on any machine.

internal/app/cli.go
```diff
@@ -191,21 +191,21 @@ func (c *cli) parse() {
...
 // readState gets the desired state from files
-func (c *cli) readState(s *state) {
+func (c *cli) readState(s *state) error {
 	// read the env file
 	if len(c.envFiles) == 0 {
 		if _, err := os.Stat(".env"); err == nil {
 			err = godotenv.Load()
 			if err != nil {
-				log.Fatal("Error loading .env file")
+				return fmt.Errorf("error loading .env file")
```
jmontroy90, line 200: Why not wrap and return the original error too? Same with L208, L230, L237, L243.

internal/app/cli.go
```diff
@@ -221,49 +221,38 @@ func (c *cli) readState(s *state) {
...
+			return fmt.Errorf("Failed to merge desired state file" + f)
 		}
 	}
 
-	s.setDefaults()
-	s.initializeNamespaces()
+	s.init() // Set defaults
 	s.disableUntargetedApps(c.group, c.target)
 
-	if len(c.target) > 0 && len(s.TargetMap) == 0 {
```
jmontroy90, line 251: In this PR, this validation now occurs later, right? Why?

internal/app/cli_test.go
```diff
@@ -7,39 +7,102 @@ var _ = func() bool {
...
+			},
 		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			if got := ToolExists(tt.args.tool); got != tt.want {
-				t.Errorf("toolExists() = %v, want %v", got, tt.want)
+			s := state{}
+			if err := tt.flags.readState(&s); err != nil {
+				t.Errorf("readState() = Unexpected error: %w", err)
```
jmontroy90, line 81: `t.Errorf` doesn't respect `%w`, I think -- `%v` instead.

jmontroy90 (commented):
It'd probably be good to have a couple more cases for multiple files, especially one TOML + one YAML with duplicates and checking the final merge.

I also see why you disabled validation, but it's weird to avoid testing it. Do you know why we even want the `skip-validation` flag in general, outside of it making testing easier? Either way, maybe we can take that `s.validate()` function outside of `s.readState()` and into `Main()` so we can conditionally call it.

internal/app/command_test.go
luisdavim: yeah but these tests already existed I've just moved them to a file where I think they make more sense since the `toolExists` method is implemented in `utils.go` and not `cli.go`

internal/app/main.go
jmontroy90: Also, this is kinda a weird check in the first place. It's checking if `s.readState()`, like, did what it's supposed to. It's almost a runtime test. If `flags` have targets and the state ends up not, it's a programming error we should cover in tests, not a runtime just-in-case kind of things. What do you think?

luisdavim: This is the way it was before I've just moved this code but I think, in this case `helmsman` just exits with `0` saying "I have nothing to do so my work is done" and that's fine...

internal/app/cli_test.go
luisdavim: yeah, good catch, I'll update.

internal/app/main.go
jmontroy90: Ah, so `s.TargetMap` is conditionally populated? e.g. `flags.target` can have targets and `s.TargetMap` can have none and that's ok?

internal/app/command_test.go
jmontroy90: Sure, but if you agree we could probably just change the check to `ls` and be done with it. 😄

luisdavim: sure but I think we have tests that actually depend on those tools...

internal/app/main.go
luisdavim: correct `s.TargetMap` is populated based on `flags.target` and/or `flags.group`

internal/app/cli.go
luisdavim: I've moved it because those cause helmsman to exit and I think it makes more sense to do that from main and I don't think it makes that much of a difference checking that before or after the syntax validation.
Now that I think about it, I think it's actually better to do it after so you learn about syntax errors in your DSF even if you made a mistake and targeted some app or group that doesn't exit, if you don't care about that you cant use `--skip-validation`...

luisdavim:
> It'd probably be good to have a couple more cases for multiple files, especially one TOML + one YAML with duplicates and checking the final merge.
> 
> I also see why you disabled validation, but it's weird to avoid testing it. Do you know why we even want the `skip-validation` flag in general, outside of it making testing easier? Either way, maybe we can take that `s.validate()` function outside of `s.readState()` and into `Main()` so we can conditionally call it.

That's a good idea, I can add more cases with the merges.
I've disabled the validation for 2 reasons:
1. the validation function has its own tests
2. (this is the main reason) validation fails with the minimum files if there's no `kubectl` context

We can add some extra test files and enable validation for those latter...

internal/app/command_test.go
jmontroy90: Oof, fair enough. 😃

jmontroy90: It's not a big deal, but maybe some day we can get rid of this.

jmontroy90:
> > It'd probably be good to have a couple more cases for multiple files, especially one TOML + one YAML with duplicates and checking the final merge.
> > I also see why you disabled validation, but it's weird to avoid testing it. Do you know why we even want the `skip-validation` flag in general, outside of it making testing easier? Either way, maybe we can take that `s.validate()` function outside of `s.readState()` and into `Main()` so we can conditionally call it.
> 
> That's a good idea, I can add more cases with the merges.
> I've disabled the validation for 2 reasons:
> 
> 1. the validation function has its own tests
> 2. (this is the main reason) validation fails with the minimum files if there's no `kubectl` context
> 
> We can add some extra test files and enable validation for those latter...

Oh okay if there's already tests for it then that's better.

jmontroy90 (approved):

PR #553, opened 2020-11-24
feat: Helm test command should run first after the release is installed.

jmontroy90:
Helm tests currently modify their Helm releases, clobbering any labels / annotations. This has previously caused problems with Helmsman's labelling, and will also cause problems if a shell post-hook does anything with those release secrets.

As such, we should run Helm tests first thing, immediately after the release is installed / upgraded.


luisdavim (approved):
LGTM

PR #554, opened 2020-11-24
feat: commands should optionally log stdout on failures.

jmontroy90:
Shell hooks potentially have helpful `stdout` contents that we'd want to optionally (verbose only) see in an error scenario.


luisdavim (approved):
Ok

PR #557, opened 2020-11-26
fix: --show-diff missing on chart version bump

### LINKED ISSUE #555: -show-diff missing on chart version bump
Hi @luisdavim, looks like these change prevent the diff from printing when using the -show-diff flag with a chart version bump, as it shortcuts the diff test: https://github.com/Praqma/helmsman/blob/master/internal/app/decision_maker.go#L323

Maybe r.diff() should run in all cases with -show-diff or -verbose mode?

luisdavim:
This should fix #555

### DIFF
--- a/internal/app/decision_maker.go
+++ b/internal/app/decision_maker.go
@@ -1,6 +1,7 @@
 package app
 
 import (
+	"fmt"
 	"regexp"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


internal/app/release.go
```diff
@@ -401,6 +397,9 @@ func (r *release) getHelmArgsFor(action string, optionalNamespaceOverride ...str
 }
 
 func (r *release) checkChartDepUpdate() {
+	if r == nil || r.disabled {
```
antoinedeschenes, line 400: Could probably use `!r.isConsideredToRun` here

internal/app/decision_maker.go
```diff
@@ -305,31 +308,43 @@ func (cs *currentState) inspectUpgradeScenario(r *release, p *plan, c *chartInfo
...
 		r.upgrade(p)
 		p.addDecision("Release [ "+r.Name+" ] will be upgraded", r.Priority, change)
-		return
+		return nil
 	}
-	if diff := r.diff(); diff != "" {
+
+	if diff, err := r.diff(); err != nil {
+		return err
+	} else if diff != "" {
```
antoinedeschenes, line 342: We're missing  `if flags.verbose || flags.showDiff {fmt.Println(diff)}` here

internal/app/release.go
luisdavim: sure.

antoinedeschenes:
lgtm, show-diff on update and upgrade works properly

antoinedeschenes (approved):

PR #566, opened 2021-01-04
Accept nil targetRelease on executing command, just omit hooks part

### LINKED ISSUE #565: Helmsman fails on deleting untracked release
For plan with:
```2021-01-04 14:04:55 NOTICE: -------- PLAN starts here --------------
2021-01-04 14:04:55 WARNING: Untracked release [ filebeat ] found and it will be deleted -- priority: -1000
```

results in

```2021-01-04 14:04:55 INFO: Printing the commands of the current plan ...
helm uninstall filebeat --namespace filebeat
2021-01-04 14:04:55 INFO: Sorting the commands in the plan based on priorities (order flags) ... 
2021-01-04 14:04:55 INFO: Executing plan
2021-01-04 14:04:55 ERROR: nil target release
```

because helmRelease uninstall command does not defines targetRelease considered hard dependency for releaseWithHooks function

mkubaczyk:
Should fix #565 

I can't remember why there was such if statement on the beginning of releaseWithHooks function so I'll wait for review.

I've just wrapped before and after hooks parts in IF statement so the main cmdExec can be invoked even if targetRelease is nil (like for removing untracked release)

### DIFF
--- a/internal/app/plan.go
+++ b/internal/app/plan.go
@@ -135,47 +135,45 @@ func releaseWithHooks(cmd orderedCommand, storageBackend string, wg *sync.WaitGr
 		wg.Done()
 		<-sem
 	}()
-	if cmd.targetRelease == nil && !flags.destroy {
-		err := fmt.Errorf("nil target release")
-		errors <- err
-		log.Verbose(err.Error())
-		return
-	}
 	var annotations []string
-	for _, c := range cmd.beforeCommands {
-		if err := execOne(c.Command, cmd.targetRelease); err != nil {
-			errors <- err
-			if key, err := c.getAnnotationKey(); err != nil {
-				annotations = append(annotations, key+"=failed")
+	if cmd.targetRelease != nil && !flags.destroy {
+		for _, c := range cmd.beforeCommands {
+			if err := execOne(c.Command, cmd.targetRelease); err != nil {
+				errors <- err
+				if key, err := c.getAnnotationKey(); err != nil {
+					annotations = append(annotations, key+"=failed")
+				}
+				log.Verbose(err.Error())
+				return
 			}
-			log.Verbose(err.Error())
-			return
+			if key, err := c.getAnnotationKey(); err != nil {
+				annotations = append(annotations, key+"=ok")
+			}
 		}
-		if key, err := c.getAnnotationKey(); err != nil {
-			annotations = append(annotations, key+"=ok")
+		if !flags.dryRun && !flags.destroy {
+			defer func() {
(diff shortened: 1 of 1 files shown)


internal/app/plan.go
```diff
@@ -135,47 +135,45 @@ func releaseWithHooks(cmd orderedCommand, storageBackend string, wg *sync.WaitGr
...
-	}
 	var annotations []string
-	for _, c := range cmd.beforeCommands {
-		if err := execOne(c.Command, cmd.targetRelease); err != nil {
-			errors <- err
+	if cmd.targetRelease != nil && !flags.destroy {
+		for _, c := range cmd.beforeCommands {
+			if err := execOne(c.Command, cmd.targetRelease); err != nil {
+				errors <- err
+				if key, err := c.getAnnotationKey(); err != nil {
```
luisdavim, line 143: I know this wasn't introduced by this PR but I think this is wrong, when `getAnnotationKey` returns a non `nil` error, the `key` will be empty, so we should only append the annotation if the error is `nil` right? Or am I missing something?

internal/app/plan.go
```diff
@@ -135,47 +135,45 @@ func releaseWithHooks(cmd orderedCommand, storageBackend string, wg *sync.WaitGr
...
+					annotations = append(annotations, key+"=failed")
+				}
+				log.Verbose(err.Error())
+				return
+			}
 			if key, err := c.getAnnotationKey(); err != nil {
-				annotations = append(annotations, key+"=failed")
+				annotations = append(annotations, key+"=ok")
```
luisdavim, line 150: same as above, we can't append the annotation if the key is empty....

internal/app/plan.go
```diff
@@ -135,47 +135,45 @@ func releaseWithHooks(cmd orderedCommand, storageBackend string, wg *sync.WaitGr
...
+			if err := execOne(c.Command, cmd.targetRelease); err != nil {
+				errors <- err
+				if key, err := c.getAnnotationKey(); err != nil {
+					annotations = append(annotations, key+"=failed")
+				}
+				log.Verbose(err.Error())
+			} else {
+				if key, err := c.getAnnotationKey(); err != nil {
+					annotations = append(annotations, key+"=ok")
+				}
```
luisdavim, line 176: same here...

luisdavim (commented):
Your changes LGTM but I think we have a problem when appending the annotations that you could also address in this PR.

luisdavim:
I've opened https://github.com/Praqma/helmsman/pull/568 to address my comments there.

PR #569, opened 2021-01-13
fix: lifecycle hook file paths not being resolved

### LINKED ISSUE #567: Issue with lifecycle hook file paths in v3.6.1+
We are having an issue upgrading to helmsman latest (v3.6.2). When running an helmsman plan (without changing anything with our working v3.4.5 deployment), helmsman fails to validate the preinstall hook files.

Our filesystem looks like this:

```
config/
    helmsman/
        helmsman-deployment.yaml
        datadog/
            values.yaml
        preinstall-hook/
           datadog/
                psp-rolebinding.yaml
```

Our app configuration for this app is 
```
datadog:
    <<: *template
    namespace: datadog
    enabled: true
    chart: datadog/datadog
    version: 2.6.3
    valuesFiles: [
      "./datadog/values.yaml",
    ]
    hooks:
      preInstall: "./preinstall-hook/datadog/psp-rolebinding.yaml"
```
_We also tried with `preInstall: "preinstall-hook/..."` and had the same result._

When running command 
```
helmsman3.6.1 -f config/helmsman/helmsman-deployment.yaml -subst-env-values
```
We get :
```
CRITICAL: apps validation failed -- for app [datadog ]. preinstall-hook/datadog/psp-rolebinding.yaml must be valid relative (from dsf file) file path
```

We've tested and this issue seems to have been introduced between v3.6.0 and v3.6.1. Testing on v3.4.5, v3.5.1 and v3.6.0 works for validating all the file paths.

luisdavim:
This should fix #567

### DIFF
--- a/internal/app/release_files.go
+++ b/internal/app/release_files.go
@@ -19,7 +19,7 @@ func (r *release) substituteVarsInStaticFiles() {
 	for key, val := range r.Hooks {
 		if key != "deleteOnSuccess" && key != "successTimeout" && key != "successCondition" {
 			hook := val.(string)
-			if err := isValidFile(hook, []string{".yaml", ".yml"}); err == nil {
+			if isOfType(hook, []string{".yaml", ".yml"}) {
 				r.Hooks[key] = substituteVarsInYaml(hook)
 			}
 		}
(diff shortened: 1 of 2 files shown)


internal/app/release_files.go
```diff
@@ -35,18 +35,18 @@ func (r *release) resolvePaths(dir, downloadDest string) {
...
+		r.SecretsFiles[i], _ = resolveOnePath(file, dir, downloadDest)
 	}
 
 	for key, val := range r.Hooks {
 		if key != "deleteOnSuccess" && key != "successTimeout" && key != "successCondition" {
-			hook := val.(string)
-			if err := isValidFile(hook, []string{".yaml", ".yml", ".json"}); err == nil {
-				r.Hooks[key], _ = resolveOnePath(hook, dir, downloadDest)
+			file := val.(string)
+			if isOfType(file, []string{".yaml", ".yml", ".json", ".sh", ".py", ".rb"}) {
```
luisdavim, line 48: Not sure if we should be checking the extensions here, we could skip the condition but that might fail if we use commands that are in the `PATH` like `kubectl` for example, any suggestions?

luisdavim:
@maxlambnc tested this and confirmed it fixes #567 (thanks for testing)

PR #574, opened 2021-01-18
MS teams notification

invingagan1:
Helmsman will send the notification to the Microsoft Teams channel. Users need to provide the MS Teams incoming connector and information to be sent in the notification.

* It is my first PR. Any suggestion is most welcome.


luisdavim (approved):
LGTM, thanks

PR #595, opened 2021-03-29
replace deprecated helm-secrets repo

antoinedeschenes:
Helm plugin is currently maintained here: https://github.com/jkroepke/helm-secrets


antoinedeschenes:
The newer plugin versions work for us, but having someone else testing this could help

luisdavim:
can you check why the tests are failing? I think you might have to also change the Circle CI config.

antoinedeschenes:
Got it. Looks like sops was installed through the deprecated plugin previously. That was just removed in the latest version of v3.

antoinedeschenes:
Should `sops` be installed with `make helmPlugins`?

luisdavim:
> Should `sops` be installed with `make helmPlugins`?

Probably yes, either in `helmPlugins` or in `tools`...

Issue #598, opened 2021-04-14
refactor: the Command struct's `Exec()` should return (ExitStatus, err)


jmontroy90:
Code that uses `ExitStatus` returned from some `status := cmd.Exec()` typically conditions on `status.code != 0`. Then `status.errors` is included in the subsequent error.

This frequently leads to information being dropped:
- Not every CLI displays anything other than a non-zero exit code for errors (often the error code itself has semantic meaning)
- Not every CLI we might use dumps errors to stderr, instead dumping to stdout
- Sometimes more relevant info for debugging is on stdout
- Slightly different: we're often constructing `ExitStatus` directly from application failures instead of invoked cmd failures, conflating the two failure modes.

As such, I propose the following refactor:

1. `cmd.Exec()` returns `(ExitStatus, err)`.
2. `ExitStatus` would be roughly the same struct, and would have a `String()` that can also be called that will include any relevant CLI stuff (stdout, stderr)
3. The returned `err` would just be some `FailureStatus` struct that implements `Error()`, where that implementation would include stdout, stderr, and the return code

The benefit of this is that we'd be able to have pretty idiomatic, consistent code everywhere, e.g.:
```
out, err := myCmd.exec()
if err != nil {
    return fmt.Errorf("error in my special command: %w", err)
}

return fmt.Sprintf("my cool cmd results: %s", out)
```

I've got a start on this already, but wanted to get some input before I committed.

luisdavim:
Sounds good to me but I don't see a reason for the error to be some custom structure, it can simply be a regular error and we'd still return the ExitStatus even on errors and carry the information there....

luisdavim:
fixed in #601

PR #601, opened 2021-04-19
feat: command.Exec() returns (*ExitStatus, error) enabling more idiomatic error management.

jmontroy90:
Per https://github.com/Praqma/helmsman/issues/598

But I ran into something that should probably be fixed first: https://github.com/Praqma/helmsman/pull/600

Based on this and that, I won't be looking to merge this until I add some (locally-passing!) tests. In particular the constant parsing of `ExitStatus.output` is a bit worrisome and merits some more tests (I saw some JSON unmarshaling!). I don't think the error exit status is parsed.

### DIFF
--- a/internal/app/command.go
+++ b/internal/app/command.go
@@ -6,7 +6,6 @@ import (
 	"math"
 	"os/exec"
 	"strings"
-	"syscall"
 	"time"
 )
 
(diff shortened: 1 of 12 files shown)


internal/app/command.go
```diff
@@ -69,33 +56,58 @@ func (c *Command) Exec() ExitStatus {
...
+		output: fmt.Sprintf(
+			"--- stdout ---\n%s\n--- stderr ---\n%s",
+			strings.TrimSpace(stdout.String()),
+			strings.TrimSpace(stderr.String())),
+	}
+}
+
+func (ee *exitError) Error() string {
+	return fmt.Sprintf("cmd %s failed with non-zero exit code %d\noutput: %s", ee.cmd, ee.code, ee.output)
+}
```
luisdavim, line 109: if you're providing a custom error type why not export it so it can be used to retrieve the error details? Otherwise this seems overkill and all we really need is `fmt.Errorf`, this is what I meant on my comment in the issue.

luisdavim (commented):
I think returning an error makes a lot of sense but as I've mentioned in the issue, I'd rather keep the error simple and just an error and get the info from the `ExitStatus`, if the problem was that before we weren't printing both `stdout` and `stderr` on errors that can be fixed without adding a new type.

luisdavim:
I've created this alternative base on your PR: https://github.com/Praqma/helmsman/pull/602

luisdavim (approved):
LGTM

PR #608, opened 2021-05-05
fix: include stdout and stderr in all Exec returns.

jmontroy90:
We should optimistically always return stdout/stderr when we `cmd.Exec()`, even if the err isn't an `*exec.ExitError`.


luisdavim (approved):
OK

PR #615, opened 2021-06-11
feat: redact --token and --password arguments

justenwalker:
When printing kubectl commands for debug logs, redact the --token and
--password arguments so that it doesn't leak credentials into logs.


luisdavim (approved):
Thanks :)

PR #619, opened 2021-08-09
Pass error to user if we fail to correctly pull or export oci chart images

Chronojam:
Hiya,

This PR just sets up helmsman to throw if its unable to save or export OCI-based charts

Thanks!


luisdavim (approved):
LGTM, thanks.

PR #626, opened 2021-08-28
build: update kubectl to 1.22.1

chenrui333:


luisdavim:
This version is already being used.

PR #627, opened 2021-09-17
feat: support the new OCI implementation from helm 3.7.0

luisdavim:
For details see:
* https://github.com/helm/helm/releases/tag/v3.7.0
* https://github.com/helm/community/blob/main/hips/hip-0006.md

### DIFF
--- a/internal/app/utils_test.go
+++ b/internal/app/utils_test.go
@@ -5,6 +5,46 @@ import (
 	"testing"
 )
 
+func TestOciRefToFilename(t *testing.T) {
+	tests := []struct {
+		name string
+		in   string
+		want string
+	}{
+		{
+			name: "no_repo",
+			in:   "my-chart:1.2.3",
+			want: "my-chart-1.2.3.tgz",
+		},
+		{
+			name: "two_colons",
+			in:   "my:chart:1.2.3",
+			want: "my:chart-1.2.3.tgz",
+		},
+		{
+			name: "with_Host",
+			in:   "my-repo.example.com/charts/my-chart:1.2.3",
+			want: "my-chart-1.2.3.tgz",
+		},
+		{
+			name: "full_url",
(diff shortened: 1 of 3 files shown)


internal/app/utils_test.go
```diff
@@ -5,6 +5,71 @@ import (
...
+			want: "my-chart-1.2.3.tgz",
+		},
+		{
+			name: "two_colons",
+			in:   "my:chart:1.2.3",
+			want: "my:chart-1.2.3.tgz",
+		},
+		{
+			name: "with_Host",
+			in:   "my-repo.example.com/charts/my-chart:1.2.3",
```
antonlisovenko, line 56: doesn't oci ref start with `oci://`? If so may adding a test case for this as well

antonlisovenko (approved):
this was very quick after Helm 3.7.0 release 🥇

internal/app/utils_test.go
```diff
@@ -5,6 +5,71 @@ import (
...
+		r  rune
+		i  int
+	}
+	tests := []struct {
+		name string
+		args args
+		want string
+	}{
+		{
+			name: "last_colon",
```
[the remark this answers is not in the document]
luisdavim: Yeah, I can actually remove the whole test since this is covered in the one below, it was just that I did this before adding the other method and then because there was coverage there I didn't add any more cases here...

internal/app/utils_test.go
luisdavim: Sure, I can do that.

luisdavim: done.

internal/app/utils_test.go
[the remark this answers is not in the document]
luisdavim: I've removed this test since the method is already covered in `TestOciRefToFilename`.

antonlisovenko (approved):

PR #633, opened 2021-11-22
Add -parent flag to support parent state files, add priority override

adamarnold-msm:
Add a -parent flag which allow for reference to multiple state files in yaml.  We find it useful to maintain state definitions within a main 'parent' file so that pipeline code does not need to build up -f flags


examples/example-parent.yaml
```diff
@@ -0,0 +1,6 @@
+--- 
+StatePaths:
```
mkubaczyk, line 2: following the naming convention from DSF, this should be `statePaths`.
I'd even consider calling it `stateFiles` instead

```suggestion
stateFiles:
```

examples/example-parent.yaml
```diff
@@ -0,0 +1,6 @@
+---
```
mkubaczyk, line 1:
```suggestion
---
```

docs/cmd_reference.md
```diff
@@ -76,6 +76,9 @@ This lists available CMD options in Helmsman:
 
   `--no-ssm-subst`
         turn off SSM parameter substitution globally.
+
```
mkubaczyk, line 79:
```suggestion

```

docs/cmd_reference.md
```diff
@@ -76,6 +76,9 @@ This lists available CMD options in Helmsman:
 
   `--no-ssm-subst`
         turn off SSM parameter substitution globally.
+ 
+  `-parent string`
+    	  parent desired state file name, contains locations of sub state files to be merged
```
mkubaczyk, line 81:
```suggestion
          parent desired state file name, contains locations of sub state files to be merged
```

internal/app/parent_state.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+type StatePaths struct {
+	// parentPath string
+	StatePaths []ParentConfig `yaml:"StatePaths"`
+}
+
+// fromYAML reads a yaml file and decodes it to a state type.
+// parser which throws an error if the YAML file is not valid.
+func (pc *StatePaths) parentFromYAML(file string) error {
```
mkubaczyk, line 21:
```suggestion
func (pc *StateFiles) parentFromYAML(file string) error {
```

tests/Invalid_example_parent.yaml
```diff
@@ -0,0 +1,4 @@
+--- 
+StatePaths:
```
mkubaczyk, line 2:
```suggestion
---
StateFiles:
```

internal/app/parent_state_test.go
```diff
@@ -0,0 +1,50 @@
+package app
+
+import (
+	"testing"
+)
+
+func Test_parentFromYAML(t *testing.T) {
+	type args struct {
+		file string
+		s    *StatePaths
```
mkubaczyk, line 10:
```suggestion
		s    *StateFiles
```

internal/app/parent_state_test.go
```diff
@@ -0,0 +1,50 @@
...
+	tests := []struct {
+		name string
+		args args
+		want bool
+	}{
+		{
+			name: "test case 1 -- Valid YAML",
+			args: args{
+				file: "../../examples/example-parent.yaml",
+				s:    new(StatePaths),
```
mkubaczyk, line 21:
```suggestion
				s:    new(StateFiles),
```

internal/app/parent_state_test.go
```diff
@@ -0,0 +1,50 @@
...
+			},
+			want: true,
+		}, {
+			name: "test case 2 -- Invalid Yaml",
+			args: args{
+				file: "../../tests/Invalid_example_parent.yaml",
+				s:    new(StatePaths),
```
mkubaczyk, line 28:
```suggestion
				s:    new(StateFiles),
```

tests/Invalid_example_parent.yaml
```diff
@@ -0,0 +1,4 @@
...
+  name1: invalid/example.yaml
+  name2: invalid/minimal-example.yaml
```
mkubaczyk, line 4:
```suggestion
  name2: invalid/minimal-example.yaml

```

internal/app/parent_state.go
```diff
@@ -0,0 +1,53 @@
...
+
+type ParentConfig struct {
+	Path     string `yaml:"path"`
+	Priority int    `yaml:"priority"`
+}
+
+type StatePaths struct {
+	// parentPath string
+	StatePaths []ParentConfig `yaml:"StatePaths"`
+}
```
mkubaczyk, line 17:
```suggestion
type StateFiles struct {
	// parentPath string
	StateFiles []ParentConfig `yaml:"stateFiles"`
}
```

internal/app/cli.go
```diff
@@ -218,32 +247,51 @@ func (c *cli) readState(s *state) error {
 	os.RemoveAll(tempFilesDir)
 	_ = os.MkdirAll(tempFilesDir, 0o755)
 
+	if len(c.parent) > 0 {
+
+		sp := new(StatePaths)
+		sp.parentFromYAML(c.parent)
+
+		for _, val := range sp.StatePaths {
```
mkubaczyk, line 255:
```suggestion
		sp := new(StateFiles)
		sp.parentFromYAML(c.parent)

		for _, val := range sp.StateFiles {
```

mkubaczyk (changes requested):
@adamarnold-msm I really appreciate your PR! I see your point, I have even went through it locally. 
I'm leaving few comments related mostly to some formatting issues or naming conventions, but despite that - LGTM.

@luisdavim please also take a look at that and consider my proposal on name change

tests/Invalid_example_parent.yaml
mkubaczyk, line 4:
```suggestion
  name2: invalid/minimal-example.yaml

```

luisdavim (commented):
I agree with all the comments @mkubaczyk made. But I'm not sure about the name of the flag, maybe something like include would be more intuitive than parent.

mkubaczyk:
@luisdavim 
how about `-spec` then? Specification file consisting of multiple desired state files, it leaves enough room for us to eventually expand the spec file in future if needed

luisdavim:
> @luisdavim how about `-spec` then? Specification file consisting of multiple desired state files, it leaves enough room for us to eventually expand the spec file in future if needed

yeah, I think that would work better than `parrent` 👍

adamarnold-msm:
Hey @mkubaczyk @luisdavim, thanks for the reviews - i've updated to use the -spec flag 👍

luisdavim (commented):
LGTM, what do you think about also allowing to specify the `stateFiles` structure in the state file so that we'd have a main desired state file that includes other files?

internal/app/spec_state.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+func checkSpecValid(f string) error {
+	if err := isValidFile(f, []string{"yaml", "yml"}); err != nil {
+		// exit
+		return err
+	}
+
+	return nil
+}
```
luisdavim, line 43: Do we really need this wrapper? And why not just:

```go
func checkSpecValid(f string) error {
	return isValidFile(f, []string{"yaml", "yml"})
}
```

internal/app/cli.go
```diff
@@ -218,32 +247,51 @@ func (c *cli) readState(s *state) error {
...
+	if len(c.spec) > 0 {
+
+		sp := new(StateFiles)
+		sp.specFromYAML(c.spec)
+
+		for _, val := range sp.StateFiles {
+			fo := fileOption{}
+			fo.name = val.Path
+			fo.priority = val.Priority
+			checkSpecValid(fo.name)
```
luisdavim, line 259: The returned error is not being checked.

internal/app/spec_state.go
```diff
@@ -0,0 +1,52 @@
...
+
+func (s *state) patchPriority(priority int) error {
+	for app, _ := range s.Apps {
+		if s.Apps[app].Priority > priority {
+			s.Apps[app].Priority = priority
+		}
+	}
+	return nil
+}
```
luisdavim, line 52: I think it would be better and more intuitive to just stable sort the list of files by priority before merging them?

internal/app/cli.go
```diff
@@ -35,7 +58,8 @@ func (i *stringArray) Set(value string) error {
 
 type cli struct {
 	debug                 bool
-	files                 stringArray
+	files                 fileOptionArray
```
luisdavim, line 61: Does this affect the existing functionality of passing multiple state files using the `-f` flag? Has this been tested with your branch?

internal/app/cli.go
```diff
@@ -218,32 +247,51 @@ func (c *cli) readState(s *state) error {
...
 
-		if err := fileState.fromFile(f); err != nil {
+		if err := fileState.fromFile(f.name); err != nil {
 			return err
 		}
-		log.Infof("Parsed [[ %s ]] successfully and found [ %d ] apps", f, len(fileState.Apps))
+
+		if f.priority != 0 {
+			fileState.patchPriority(f.priority)
+		}
```
luisdavim, line 274: This might have some edge cases and I think it would be better to stable sort the list of files based on their priority before this loop.

mkubaczyk:
@adamarnold-msm, I have prepared the changes based on @luisdavim review and created PR to your branch: https://github.com/adamarnold-msm/helmsman/pull/1
You can merge it so we can review current PR :-)

luisdavim (commented):
Thanks for the updates, now that the files are sorted based on their priority I don't think we need the `patchPriority` function any more, the apps will be added in the same order as they appear in the files, if they set a priority that should be respected.

I think that keeping it might confuse users if the priorities they set are overwritten. The file priority should only influence the order the different spec files are parsed, not the relative priority of the apps, those are different things, users might wan to use a higher priority file to lower the priority of some app by overwriting it.

adamarnold-msm:
@luisdavim, I'll remove patchPriority() and test that today

internal/app/cli.go
mkubaczyk: I've checked that when adding my changes, worked as it was before

mkubaczyk (approved):

mkubaczyk:
> What do you think about also allowing to specify the `stateFiles` structure in the state file so that we'd have a main desired state file that includes other files?

by that you meant having `stateFiles` in DSF on the same level as `settings` or `namespaces` blocks are?

luisdavim:
> > What do you think about also allowing to specify the `stateFiles` structure in the state file so that we'd have a main desired state file that includes other files?
> 
> @luisdavim by that you meant having `stateFiles` in DSF on the same level as `settings` or `namespaces` blocks are?

Yes, something like an include section. But that's better for a separate PR after this is merged.

mkubaczyk:
@luisdavim yes, that's the same I wanted to propose, separate PR. Current state of this PR LGTM. What we're missing is just the docs but we can add them independently before doing a release

luisdavim:
@adamarnold-msm why did you remove the priority? I think we do need to be able to set a priority on the files in the case the user needs to control the order they are processed and merged into the final confi.

adamarnold-msm:
> @adamarnold-msm why did you remove the priority? I think we do need to be able to set a priority on the files in the case the user needs to control the order they are processed and merged into the final config.

Hey @luisdavim, i tested this last night, the sorting now is on the priority in fileOption

type fileOption struct {
 	name     string
 	priority int
 }

instead of specifying a priority in the spec file, or have i misunderstood?  Thanks

luisdavim (approved):

PR #634, opened 2021-12-03
Add skipIgnoredApps option to DSF Settings block

mkubaczyk:
Adding `skipIgnoredApps` option in Settings block of DSF file.
When defined as `true`, output plan will be limited to the apps with decision different than `ignored`. This will help in multiple helmsman runs when `-target` or `-group` flag is used on DSF having significant number of apps causing the plan to contain tons of `ignored` lines making the plan hard to read.
What do you think?

### DIFF
--- a/internal/app/decision_maker.go
+++ b/internal/app/decision_maker.go
@@ -77,7 +77,7 @@ func (cs *currentState) makePlan(s *state) *plan {
 				<-sem
 			}()
 			r.checkChartDepUpdate()
-			cs.decide(r, s.Namespaces[r.Namespace], p, c)
+			cs.decide(r, s.Namespaces[r.Namespace], p, c, s.Settings)
 		}(r, s.ChartInfo[r.Chart][r.Version])
 	}
 	wg.Wait()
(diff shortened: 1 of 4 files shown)


internal/app/decision_maker.go
```diff
@@ -87,10 +87,12 @@ func (cs *currentState) makePlan(s *state) *plan {
 
 // decide makes a decision about what commands (actions) need to be executed
 // to make a release section of the desired state come true.
-func (cs *currentState) decide(r *release, n *namespace, p *plan, c *chartInfo) {
+func (cs *currentState) decide(r *release, n *namespace, p *plan, c *chartInfo, settings config) {
 	// check for presence in defined targets or groups
 	if !r.isConsideredToRun() {
-		p.addDecision("Release [ "+r.Name+" ] ignored", r.Priority, ignored)
+		if !settings.SkipIgnoredApps {
```
luisdavim, line 93: nit: can you use `&&` instead of nesting?

internal/app/decision_maker.go
```diff
@@ -87,10 +87,12 @@ func (cs *currentState) makePlan(s *state) *plan {
 
 // decide makes a decision about what commands (actions) need to be executed
 // to make a release section of the desired state come true.
-func (cs *currentState) decide(r *release, n *namespace, p *plan, c *chartInfo) {
+func (cs *currentState) decide(r *release, n *namespace, p *plan, c *chartInfo, settings config) {
```
luisdavim, line 90: nit: do we really need to pass the whole settings here or should we just pass the skip ignore flag?, Not a big deal since this is a long running app, just thinking out loud.

luisdavim (approved):
LGTM, just a couple of comments.

internal/app/decision_maker.go
mkubaczyk: I don't think `and` here will do the job as we're deciding on adding or not adding the decision in two states of `SkipIgnoredApps` but the return must happen in both cases

internal/app/decision_maker.go
mkubaczyk: I did that assuming some more fields can be used from settings inside the function. Do you think it's valid assumption to leave it as I did it?

internal/app/decision_maker.go
luisdavim: You're right, I missed that.

internal/app/decision_maker.go
luisdavim: Yeah, let's keep it as you have it here.

PR #638, opened 2022-01-21
Fix broken links in examples

tjg184:
Links were incorrect and getting 404s for `example.yaml` and `example.toml`.  Looks like they might have moved at some point.


luisdavim (approved):
Thanks for spotting that.

PR #640, opened 2022-03-02
Option for checking for available updates for app charts

weeezes:
Hi!

Here's a small feature for running a check for available updates for charts used in the states apps.

It's often easier to keep up with the chart updates when moving in smaller steps, but keeping up on what could be updated is an annoying chore. This feature helps a bit with the tracking.


luisdavim (approved):
LGTM, tanks.

PR #642, opened 2022-03-21
Support arm64 builds

mixja:
This PR updates the cross compilation task to add support for Linux/MacOS arm64 builds.

With this PR the following builds are created:

```
% make cross
Number of parallel builds: 9

-->    darwin/amd64: github.com/Praqma/helmsman/cmd/helmsman
-->     linux/arm64: github.com/Praqma/helmsman/cmd/helmsman
-->   windows/amd64: github.com/Praqma/helmsman/cmd/helmsman
-->    darwin/arm64: github.com/Praqma/helmsman/cmd/helmsman
-->     linux/amd64: github.com/Praqma/helmsman/cmd/helmsman
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,7 +77,7 @@ test: deps vet repo ## Run unit tests
 .PHONY: test
 
 cross: deps ## Create binaries for all OSs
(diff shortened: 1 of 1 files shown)


mkubaczyk:
it sounds ok to me, but at this point we can't be sure there's no one using some other architecture build so I hope this will be raised if someone misses that after this is merged :-)

Issue #639, opened 2022-02-02
Helmsman performance using OCI registries


Revellski:
Helmsman seems to work considerably slower when using OCI registries. Has anyone else noticed similar issues?

Using ChartMuseum registry a single run averages ~1m45s. Using OCI registry (following [this doc](https://github.com/Praqma/helmsman/blob/4846415ebf12f88894db2401728dbde2014d0edf/docs/how_to/helm_repos/oci.md)) it takes ~4m52s for the same setup (releases, versions & etc).

Tweaking the concurrency using `-p int` from 1 to 15 (considering there are 15 charts) improves the run to 4m7s but is still over x2 slower than using ChartMuseum registries.

Could there be additional configuration I am missing which could improve performance? Perhaps define the OCI registries in `helmRepos` stanza? 

```
Helmsman v3.8.0
Helm v3.7.2
Helm Diff v3.1.3
Kubectl v1.20.5
```

_Update (2022/02/04):_
The debug logs reveal `helm pull oci://` is the slowest out of all operations. Unfortunately, it does not respect the concurrency settings passed by the `-p {int}` parameter but appears to be synchronous.

luisdavim:
We can look into adding parallelization to that step, but I'm not sure if helm will allow it.

luisdavim:
I'm closing this as I'm pretty sure this is no longer an issue with newer versions of `hlem` and `helmsman` feel free to re-open if you find it's still an issue.

PR #643, opened 2022-03-24
feat: don't download charts if helm can do it by itself

luisdavim:
I was looking into #639 and it occurred to me that since `helm v3.8.0` we don't really need to pre-download `oci` charts and that should make things faster, I've added a flag `--download-charts` to allow forcing the old behaviour where all charts referenced by a URL would be downloaded.


mkubaczyk (approved):

mitchellhuang:
Just a heads up that this PR is great, but also surfaced a helm bug for us with `--registry-config` or environment variable causing the helm install of OCI charts to fail. Looks like it will be fixed in helm in 3.10. To mitigate for the time being we have used the flag `--download-charts`, thanks for adding this flag!

https://github.com/helm/helm/pull/10762

PR #644, opened 2022-03-27
State composition

luisdavim:
Just some refactoring and adding more tests.
Each commit can be reviewed separately.


mkubaczyk (approved):
looks good, I have run few single tests and it worked fine

PR #646, opened 2022-04-03
Pending states

luisdavim:
Allow ignoring apps in pending states or try to wait for the state to change.

### DIFF
--- a/internal/app/decision_maker.go
+++ b/internal/app/decision_maker.go
@@ -77,7 +77,9 @@ func (cs *currentState) makePlan(s *state) *plan {
 				<-sem
 			}()
 			r.checkChartDepUpdate()
-			cs.decide(r, s.Namespaces[r.Namespace], p, c, s.Settings)
+			if err := cs.decide(r, s.Namespaces[r.Namespace], p, c, s.Settings); err != nil {
+				log.Fatal(err.Error())
+			}
 		}(r, s.ChartInfo[r.Chart][r.Version])
 	}
 	wg.Wait()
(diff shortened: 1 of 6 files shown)


mkubaczyk:
How about adding one more option like `waitForPendingApps` (or something similar) as int of total seconds to wait for them to become ready (so it has to be properly addressed with `retries` loop) and if it's not defined, there will be no retries so it will immediately print the pending error message assuming `skipPendingApps` was `false`?

luisdavim:
> How about adding one more option like `pendingAppsWaitSeconds` (or something similar) as int of total seconds to wait for them to become ready (so it has to be properly addressed with `retries` loop) and if it's not defined, there will be no retries so it will immediately print the pending error message assuming `skipPendingApps` was `false`?

Yeah, I can add something like that like a max retry count or a time and I calculate the max retries from the provided time and if you set it to 0 it just fails immediately.

luisdavim:
I've added a `--pending-max-retries` flag that defaults to 0 (keeping the original behaviour).

internal/app/decision_maker.go
```diff
@@ -87,68 +91,81 @@ func (cs *currentState) makePlan(s *state) *plan {
...
+		p.addDecision(prefix+" is in FAILED state. Upgrade is scheduled!", r.Priority, change)
 		r.upgrade(p)
-		return
+		return nil
 
 	case helmStatusPendingInstall, helmStatusPendingUpgrade, helmStatusPendingRollback, helmStatusUninstalling:
-		log.Fatal("Release [ " + r.Name + " ] in namespace [ " + r.Namespace + " ] is in a pending (install/upgrade/rollback or uninstalling) state. " +
-			"This means application is being operated on outside of this Helmsman invocation's scope." +
-			"Exiting, as this may cause issues when continuing...")
+		if settings.SkipPendingApps {
```
mkubaczyk, line 143: as you are using `if settings.SkipPendingApps` condition here, is it worth passing another argument `retries` to decide()? I think you can leave decide() as it was when it comes to arguments and just use `settings.pendingAppRetries` here for retries calculation

mkubaczyk (changes requested):
hey, I left one comment regarding decide() retries argument

internal/app/decision_maker.go
luisdavim: You mean modify the value from the settings? I think the problem with that is that it is global and we call the decision function for each release and they would after each other. 

Or maybe I'm not getting what you are proposing.

mkubaczyk: no, I just meant "why do we even have to pass `retries` to decide()" when we can just use settings.pendingAppRetries in that logic's part responsible for doing a retry when pending state is encountered :-)

luisdavim: It's using recursion, so I don't have to duplicate the logic to check the state and decide, it starts by passing the max number from `settings.pendingAppRetries` but this is passed by copy so each instance of `decide` has its own value, then the `reties` argument is decremented and `decide` calls itself.

We run `decide` for each release, sometimes even in parallel, if I was to use the global `settings.pendingAppRetries` and decrement that it would only work for one of the instances. We pass it as an argument, so each instance has its own value and we can keep track of how many reties that instance has attempted.

I can do this without recursion with a loop inside the `decide` function but I'd have to duplicate some code from the top of the function into that loop, that's why I decided to go with recursion.

mkubaczyk: I see. Ok then. :-)

mkubaczyk (approved):

Issue #645, opened 2022-03-31
Env variable substitution : .bashrc/current session env takes precedence over supplied .env file


claireleray-poclain:
Hi!

It seems that when applying a configuration using a command looking like  `helmsman --apply --subst-env-values -e <config-file.env>  -f helmsman.yaml`, Helmsman first uses variables from .bashrc instead of the supplied config file. Is that the expected behavior ? If so, could that maybe made clearer in the documentation ?

Thanks!

luisdavim:
I'm not sure if I get what you are reporting. 

Any process that you start will "see" the variables that are already set in your session, for example, the ones from `bashrc`.

In this case, helmsman will read the env file you provided and load them on top of the ones that already existed before, so the variables set on the env file should take precedence. If the same variable is set both on your `bashrc` file and in the env file you provided, helmsman should end up using the value set on the env file.

Are you saying that if a variable is set on both, helmsman is ignoring the value from the env file and using the one from `bashrc`? That should not happen and if it does it's a bug.

claireleray-poclain:
Yes ! Variables were set on both, and I had to strip my `bashrc` so that Helmsman would use the ones in my env file.

luisdavim:
Ok, that should not happen. I'll have a look.

luisdavim:
This is fixed in #649 I'll cut a release soon.

PR #649, opened 2022-04-23
fix: .env files should overwrite env vars

luisdavim:
This solves #645 and makes the loading of `.env` files behave as I would expect them to but it does change the existing behaviour quite significantly:

* Before the default `.env` file would only be loaded if no env files were explicitly passed through the `-e` flag, now it will always be loaded first if present
* Before loading env files would not overwrite any env variable that had already been set before, now it does so when loading multiple files if a variable is set more than once the value from the last file to be loaded will take precedence.
* Before the first file would take precedence, now, the last one will.

I've checked the documentation and there's no mention to what is the expected behaviour.

I think this behaviour is more natural and intuitive but what do you think @mkubaczyk

### DIFF
--- a/docs/cmd_reference.md
+++ b/docs/cmd_reference.md
@@ -36,7 +36,8 @@ This lists available CMD options in Helmsman:
         apply the dry-run (do not update) option for helm commands.
 
   `-e value`
-        file(s) to load environment variables from (default .env), may be supplied more than once.
+        additional file(s) to load environment variables from, may be supplied more than once, it extends default .env file lookup, every next file takes precedence over previous ones in case of having the same environment variables defined.
+        If a `.env` file exists, it will be loaded by default, if additional env files are specified using the `-e` flag, the environment file will be loaded in order where the last file will take precedence.
 
   `-f value`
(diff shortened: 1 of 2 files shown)


luisdavim:
Mayb I could instead add a `-E` flag that does the overwrides and lead the `-e` as is but in that case waht should we do about the `.env` file? Should we ignore it if `-e` is used and load it first if `-E` is used or none is used? That could be confusing to the users...

mkubaczyk (approved):

mkubaczyk:
How about changing the `-e` description to be "additional file(s) to load environment variables from, may be supplied more than once, it extends default .env file lookup, every next file takes precedence over previous ones in case of having the same environment variables defined" or something similar? This way we'll put it explicit that .env file is being loaded no matter what and that every next file being passed has higher priority over the previous ones.

luisdavim:
Yeah, I can update the docs to make it more explicit.

luisdavim:
I've updated the flag doc string and the command reference doc.

mkubaczyk:
LGTM, now it should be clear for everyone what's the relation between all of it :-)

PR #659, opened 2022-05-19
Add -exclude-target and -exclude-group cli options

mkubaczyk:
This PR adds `-exclude-target` and `-exclude-group` CLI flags so one can exclude specific apps by either their names or groups from execution no matter if `-target` and `-group` flags were used or not. This came useful for me when trying to redeploy whole environment without one or two apps that I had to take care differently.
It works for every possible case as I'm disabling those apps on an early stage even before considering `-target` or `-group` flags.


luisdavim (approved):
LGTM

PR #661, opened 2022-05-20
Add error handling to state building from files

w3irdrobot:
I noticed that an invalid application configuration didn't error out when creating a plan. Instead it would just exit and say that no applications were found. Specifically, I had `preInstall` under the application instead of under `my-app.hooks`. Digging in, it appears that errors from parsing were being returned, but the `build` method on the state wasn't handling the error being return. This simply handles the error and returns it so that it is correctly outputted to the screen so the issue is shown to the user instead of being swallowed.


luisdavim (approved):
LGTM.

Thanks

PR #673, opened 2022-06-05
fix: check charts in all previous helmRepos

luisdavim:
Alternative implementation for #669 

I've also modified the tests to make sure this case is covered.

cc: @bergerx

### DIFF
--- a/examples/composition/argo.yaml
+++ b/examples/composition/argo.yaml
@@ -13,9 +13,6 @@ namespaces:
         - name: "requests.nvidia.com/gpu"
           value: "2"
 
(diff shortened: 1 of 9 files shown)


bergerx (approved):
Using `state.PreconfiguredHelmRepos` will likely impact re-generating the state file using the `state.toFile` function. But currently `state.toFile` is not used anywhere. This could be thought on the day we want to use that function for whatever reason.

luisdavim:
This is just merging the repository and preconfigured repos ahead of calling expand so they are all available, it won't affect the final state.

Issue #576, opened 2021-01-20
JSON schema


sacesare:
Please, provide [json schema](https://www.schemastore.org/) with comments(docs), instead of [desired state specification](https://github.com/Praqma/helmsman/blob/master/docs/desired_state_specification.md )  and some particular tom/yaml cases, for standardized IDE support/documentation.
VScode , for example, can use - [Even Better TOML](https://taplo.tamasfe.dev/configuration/#configuration-file)
Something like [wetzel](https://github.com/CesiumGS/wetzel) can be used to generate md docs from schema if required.
Thanks.

luisdavim:
Hi @sacesare I've opened #675 to add a generated json schema file, can you check if that file fits your use cases?

PR #675, opened 2022-06-11
Json schema

luisdavim:
Adds a json schema generator for the desired state definition.
Sorry for the large PR, I had to export our types and add comments to their fields to get a meaningful `jsonschema` file being generated.

Fixes: #576

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,19 +60,22 @@ update-deps: ## Update depdendencies. Runs `go get -u` internally.
 	@GOFLAGS="" go mod tidy
 	@GOFLAGS="" go mod vendor
 
-build: deps vet ## Build the package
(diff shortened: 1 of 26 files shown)


mkubaczyk (approved):

luisdavim:
> There's been tons of syntax changes so I hope I caught those meaningful parts and understood them fully. LGTM

Thanks, the real changes are in the second commit, the other one is mainly capitalizing the struct names.
There's also a change that I broke `flag.parse` into 2 separate functions where one remains being called from `init` and the other is called from `main`, this was to avoid strange side-effects whem importing the internal package for generatign the schema.

Anyay the tests are all passing and I've also done some manual testing just to be sure and it looks ok.

PR #678, opened 2022-06-24
Make -target consistent with app release name

### LINKED ISSUE #648: Helmsman cannot remove release with target flag
Helmsman cannot remove release in cluster under following conditions:
* target flag is used (with yaml key as a value)
* release name (`name` field) is different then yaml key

eg.
```
apps:
  ldap-gateway:
    enabled: false
    wait: true
    timeout: 300
    namespace: "apps"
    chart: "charts/X"
    version: "0.2.0"
    name: "ldap"
    secretsFiles:
      - "X.yaml"
```

when run with `-target ldap-gateway` it does not find any release and do nothing.

mkubaczyk:
Planning to fix #648


luisdavim (approved):
LGTM

PR #688, opened 2022-07-21
Get kubectl version by taking json output instead of parsing string

### LINKED ISSUE #687: Dry Run stopped working (--server-dry-run again)
Helm version: v3.9.1
Helmsman version: v3.12.0
kubectl version: v1.24.3

Debug log:

    2022-07-19 11:07:54 NOTICE: Apply .helmsman-tmp/tmp2026425016/monitoring-nonprod.yaml manifest postUpgrade
    2022-07-19 11:07:54 DEBUG: kubectl apply -n proxies -f .helmsman-tmp/tmp2026425016/monitoring-nonprod.yaml --server-dry-run
    2022-07-19 11:07:54 ERROR: command for release [proxy-staging] failed: Apply .helmsman-tmp/tmp2026425016/monitoring-nonprod.yaml manifest postUpgrade failed with non-zero exit code 1: exit status 1
    output:
    --- stdout ---
    
    --- stderr ---
    error: unknown flag: --server-dry-run
    See 'kubectl apply --help' for usage.
    2022-07-19 11:07:54 CRITICAL: Plan execution failed

Dry run seems to be failing, because the `--server-dry-run` flag is given to `kubectl` instead of `--dry-run=server`. Looking at the code, that shouldn't happen for recent kubectl versions, but I think the kubectl version detection in kube_helpers.go broke, because the output from kubectl has changed. For 1.24.3, the full output of `kubectl version --short --client` is

    Flag --short has been deprecated, and will be removed in the future. The --short output will become the default.
    Client Version: v1.24.3
    Kustomize Version: v4.5.4

The code at https://github.com/Praqma/helmsman/blob/19b36d15830f63351c54e67928500269edd48baf/internal/app/kube_helpers.go#L348 splits this by `:` with 2 elements, which would evaluate to

     v1.24.3
    Kustomize Version: v4.5.4

so now an additional line ends up in the `version` variable.

I'd happily provide a PR for a fix, but I'm not sure what would be the best approach that would work on enough versions of kubectl. The proper way for the future would probably be to use `kubectl version -o json` and then unmarshalling the json, but I'm not sure if that works on older versions of kubectl, and how many versions should be supported.

mkubaczyk:
I'm updating the `getKubectlVersion` function unmarshall json output of `kubectl version` instead of parsing string output. I've checked if this is supported for the command back to v1.21 of kubectl - it is.
Should fix #687

### DIFF
--- a/internal/app/kube_helpers.go
+++ b/internal/app/kube_helpers.go
@@ -337,17 +337,25 @@ func getReleaseContext(releaseName, namespace, storageBackend string) string {
 
 // getKubectlVersion returns kubectl client version
 func getKubectlVersion() string {
-	cmd := kubectl([]string{"version", "--short", "--client"}, "Checking kubectl version")
+	var kubectlVersion struct {
+		ClientVersion map[string]string `json:"clientVersion"`
+	}
 
-	res, err := cmd.Exec()
+	cmd := kubectl([]string{"version", "--output=json", "--client"}, "Checking kubectl version")
+	ver, err := cmd.Exec()
 	if err != nil {
 		log.Fatalf("While checking kubectl version: %v", err)
 	}
-	version := strings.TrimSpace(res.output)
-	if !strings.HasPrefix(version, "v") {
-		version = strings.SplitN(version, ":", 2)[1]
+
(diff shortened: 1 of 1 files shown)


luisdavim (approved):
👍

PR #699, opened 2022-08-09
Add support for helm-vault plugin

bbetter173:
This PR adds support for the [helm-vault plugin](https://github.com/Just-Insane/helm-vault) and allows most of the options to be specified as you'd expect.

I haven't written any tests that cover this behaviour, as I'm not really sure how to go about mocking something like vault in this context. If you have any ideas I'd be happy to update the PR.

### DIFF
--- a/internal/app/utils.go
+++ b/internal/app/utils.go
@@ -482,6 +482,27 @@ func decryptSecret(name string) error {
 		if settings.EyamlPrivateKeyPath != "" && settings.EyamlPublicKeyPath != "" {
 			args = append(args, []string{"--pkcs7-private-key", settings.EyamlPrivateKeyPath, "--pkcs7-public-key", settings.EyamlPublicKeyPath}...)
 		}
+	} else if settings.VaultEnabled {
+		args = []string{"vault", "dec"}
+		if settings.VaultDeliminator != "" {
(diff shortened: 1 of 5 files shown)


luisdavim:
Thanks for your contribution @bbetter173 can you update the JSON schema by running `make schema`?

internal/app/utils.go
```diff
@@ -482,6 +482,27 @@ func decryptSecret(name string) error {
 		if settings.EyamlPrivateKeyPath != "" && settings.EyamlPublicKeyPath != "" {
 			args = append(args, []string{"--pkcs7-private-key", settings.EyamlPrivateKeyPath, "--pkcs7-public-key", settings.EyamlPublicKeyPath}...)
 		}
+	} else if settings.VaultEnabled {
+		args = []string{"vault", "dec"}
```
luisdavim, line 486: This will fail if the vault plugin is missing and the resulting error might not be easy to understand, we should check if the plugin ins installed during the DSF validation if `settings.VaultEnabled` is set. You can use the `helmPluginExists` function for this.

bbetter173: Pushed some commits that should resolve these issues, thanks for the feedback.

luisdavim (approved):
LGTM, thanks.

PR #713, opened 2022-09-14
kube_helpers.go: don't look for context when SQL backend is used

g-linville:
Helm supports using a PostgreSQL database as the storage backend, rather than k8s Secrets or ConfigMaps. This is the storage backend that I use, and when I try to use Helmsman with it, I get an error because the kubectl command in this `getReleaseContext` function assumes that either Secrets or ConfigMaps are in use.

My change makes it so that this context check is altogether skipped and the default context is used if the storage backend is set to `sql`. My workaround up to this point has been to create a placeholder secret in each namespace that matches the label selectors that this function uses, but I would rather not have to do that.

### DIFF
--- a/internal/app/kube_helpers.go
+++ b/internal/app/kube_helpers.go
@@ -322,15 +322,18 @@ func getReleaseContext(releaseName, namespace, storageBackend string) string {
 	// kubectl get secrets -n test1 -l MANAGED-BY=HELMSMAN -o=jsonpath='{.items[0].metadata.labels.HELMSMAN_CONTEXT}'
 	// kubectl get secret sh.helm.release.v1.argo.v1  -n test1  -o=jsonpath='{.metadata.labels.HELMSMAN_CONTEXT}'
 	// kubectl get secret -l owner=helm,name=argo -n test1 -o=jsonpath='{.items[-1].metadata.labels.HELMSMAN_CONTEXT}'
-	cmd := kubectl([]string{"get", storageBackend, "-n", namespace, "-l", "owner=helm", "-l", "name=" + releaseName, "-o", "jsonpath='{.items[-1].metadata.labels.HELMSMAN_CONTEXT}'"}, "Getting Helmsman context for [ "+releaseName+" ] release")
+	rctx := defaultContextName
+	if storageBackend != "sql" {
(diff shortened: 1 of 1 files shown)


internal/app/kube_helpers.go
```diff
@@ -322,15 +322,18 @@ func getReleaseContext(releaseName, namespace, storageBackend string) string {
 	// kubectl get secrets -n test1 -l MANAGED-BY=HELMSMAN -o=jsonpath='{.items[0].metadata.labels.HELMSMAN_CONTEXT}'
 	// kubectl get secret sh.helm.release.v1.argo.v1  -n test1  -o=jsonpath='{.metadata.labels.HELMSMAN_CONTEXT}'
 	// kubectl get secret -l owner=helm,name=argo -n test1 -o=jsonpath='{.items[-1].metadata.labels.HELMSMAN_CONTEXT}'
-	cmd := kubectl([]string{"get", storageBackend, "-n", namespace, "-l", "owner=helm", "-l", "name=" + releaseName, "-o", "jsonpath='{.items[-1].metadata.labels.HELMSMAN_CONTEXT}'"}, "Getting Helmsman context for [ "+releaseName+" ] release")
+	rctx := defaultContextName
+	if storageBackend != "sql" {
```
luisdavim, line 326: instead of wrappint the whole code in the condition, can you just exit early?

luisdavim:
one thing to note is that not havint a way of chcking the context can be dangerous, if you have multiple helm releases in the same namespace being managed through separate helmsman configs, if they are all on the same context (default) helmsman will remove the releases it finds in the cluster that are missing from the config, unless you tell it to keep untracked releases and in that case you won't have a way to automatically remove releases when you remove them from the config...

g-linville:
> one thing to note is that not havint a way of chcking the context can be dangerous, if you have multiple helm releases in the same namespace being managed through separate helmsman configs, if they are all on the same context (default) helmsman will remove the releases it finds in the cluster that are missing from the config, unless you tell it to keep untracked releases and in that case you won't have a way to automatically remove releases when you remove them from the config...

Yeah it would definitely be bad if someone got burned by that. Do you think it would be a good idea to extend the settings to provide information on connecting to PostgreSQL, so that I can add to this function the code needed to connect to PostgreSQL and retrieve context information? That would probably require using a library like [pq](https://pkg.go.dev/github.com/lib/pq), which would add another dependency, but I would love to work on this.

luisdavim:
Yes, or at least print a warning for now and then we figure out a better way to protect this.

PR #746, opened 2022-12-20
allow use of metadata server for gcs authentication inside GCP

chrisSawarzynski:
Currently helmsman requires certificate provided via GOOGLE_APPLICATION_CREDENTIALS or GCLOUD_CREDENTIALS environment variable to read charts from GCP.

When running inside GCP helmsman can use metadata server to authenticate as Service Account.

This change prevents helmsman from failing when no env variables are provided, but metadata server is available.

### DIFF
--- a/internal/gcs/gcs.go
+++ b/internal/gcs/gcs.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"net/http"
 	"os"
 
 	// Imports the Google Cloud Storage client package.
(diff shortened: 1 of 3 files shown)


internal/gcs/gcs.go
```diff
@@ -15,8 +16,19 @@ import (
 // colorizer
 var style aurora.Aurora
 
+func IsRunningInGCP() bool {
+	resp, err := http.Get("http://metadata.google.internal")
+	if err != nil {
+		return false
+	}
+	defer resp.Body.Close()
```
luisdavim, line 24: there's no need to defer this call here...

luisdavim (commented):
Overall, LGMT just left a minor comment in the code.
Thanks

internal/gcs/gcs.go
chrisSawarzynski: changed, thanks :)

luisdavim (approved):
Thanks 👍

PR #759, opened 2023-01-24
Fix helm-vault secret decryption

### LINKED ISSUE #758: Vault decrypted secrets fail in various ways depending on version of helm-secrets installed
When using the following DSF settings:

```
settings:
  vaultEnabled: true
  vaultEnvironment: demo
```

Helmsman fails to decrypt secrets in different ways:

**Without helm-secrets >=4.0.0**

```
CRITICAL: Error decryption failed: Decryption Done
```
**With helm-secrets >= 4.0.0**

Silent failure resulting in missing values

This is due to some logic errors in the `decryptSecret` function.

bbetter173:
Fixes #758 

- `helm-vault` does not support outputting secrets to stdout, so `useHelmOutput` is set to `false`
- `helm-vault` interpolates the environment variable name into the output file if one is specified, so an additional step of renaming the file to the expected name is added
- `helm-vault` outputs the expected file by default (original file name with `.dec` appended) if no environment is specified. No changes required to handle this case other than settings `useHelmOutput` to `false`

### DIFF
--- a/internal/app/utils.go
+++ b/internal/app/utils.go
@@ -526,6 +526,19 @@ func decryptSecret(name string) error {
 		outfile += ".dec"
 	}
 
+	if settings.VaultEnabled {
+		// helm-vault plugin doesn't write to stdout
+		useHelmOutput = false
+		if settings.VaultEnvironment != "" {
+			// helm-vault decryption with an environment interpolate the environment name into the output filename
+			vaultOutFile := name + "." + settings.VaultEnvironment + ".dec"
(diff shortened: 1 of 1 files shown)


luisdavim (approved):
LGTM, thanks.

bbetter173:
@luisdavim - Is there any ETA on a release that will include this fix?

luisdavim:
I can look into cutting a release today.

bbetter173:
Thanks mate

PR #776, opened 2023-03-25
Add support for hiera-eyaml-gkms

rflj:
I have noticed, that helmsman supports `hiera-eyaml` with only local PKCS7 keys for secrets. This PR adds support for Google Cloud KMS in `hiera-eyaml`.

Currently, helmsman only supports `hiera-eyaml` with local PKCS7 keys for secrets. This limits users who want to manage their keys differently.

I have made a code changes to add support for `hiera-eyaml-gkms`, which is a plugin for `hiera-eyaml `that extends default `hiera-eyaml` functionalities. This will allow users to encrypt and decrypt secrets using GKMS keys stored in GCP. Users can specify the project id, keyring and crypto-key they are referring to.

These change will add a valuable feature to helmsman allowing users to use different method and can be a starting point to include also different cloud providers for `heira-eyaml`.


luisdavim (approved):
LGTM, thanks.

Issue #780, opened 2023-03-30
Priority cannot be configured for files in spec


Sheludchenko:
I tried to split a long file into multiple ones. I used this document **docs/how_to/misc/multiple_desired_state_files_specification.md** as a reference.

Here's a sample spec file I'm trying to use:
```
---
stateFiles:
 - path: common.yaml
   priority: -20
 - path: crd.yaml
   priority: -10
 - path: service.yaml
   priority: 0
```

Here's a command I'm trying to run:
```
helmsman --spec environment.spec.yaml --dry-run --no-fancy
```

Here's an error I'm getting:
```
2023-03-30 14:42:55 CRITICAL: error parsing spec file: yaml: unmarshal errors:
  line 4: field priority not found in type app.StatePath
```

I looked through **internal/app/spec_state.go** and didn't find any reference for **priority** field.

mkubaczyk:
Hey,

first of all, which Helmsman version are you using? When I run it for the current master state, it just ignores "priority:" field without an error.
And yes, priority is not supported. -spec file was introduced in 3.8.0, but 3.8.1 removed priority field (breaking change introduced as hotfix in versioning, that's our fault). Files are being processed in the order they are defined in the spec file from the top to bottom.

Sheludchenko:
@mkubaczyk currently I'm using Helmsman v3.11.0. Thanks for quick update.

mkubaczyk:
ok, quite an old one. I think newer versions would just ignore this priority field (can't tell if that's better or worse atm). Nevertheless priority field is deprecated and the ordering decides. :-)

PR #781, opened 2023-03-31
Remove 'priority' field from -spec docs

mkubaczyk:
Removing sort() function of fileOptionArray and "priority" part from multiple -spec file docs

Relates to #780


luisdavim (approved):
LGTM

PR #793, opened 2023-05-24
Optioned recursive environment variables expansion

tonyvelichko:
This MR will add an option to disable the recursive variables expansion. 
The default behavior is kept the same. 
CLI documentation is updated according to the changes.


luisdavim (approved):
LGTM, thanks.

PR #801, opened 2023-06-15
Add additional property required to use hiera-eyaml-gkms

rflj:
Resolved conflicts from previous PR.
Adding additional property required for hiera-eyaml-gkms


luisdavim (approved):
Thanks

PR #814, opened 2023-08-03
Fix issue with hiera-eyaml-gkms encrypt/decrypt actions

rflj:
I found issue when encrypting/decrypting secrets with hiera-eyaml using GKMS in docker container. eyaml fails with error message:
```
Error loading shared library ld-linux-x86-64.so.2
```

Issue resolved by adding `gcompat` package.


luisdavim (approved):
LGTM, thanks.

PR #874, opened 2024-02-21
Update README.md v3.17.0

Michielvk:
Update installation commands to use the latest version. Version 3.11.0 does not work anymore with the latest versions of kubectl due to the deprecation of the `--short` flag on `kubectl version`.

Maybe this can be automated in the future using a GitHub action?


luisdavim (approved):
Thanks

PR #882, opened 2024-03-16
feat: add show-secrets option

ericchu94:
Add the ability to show secrets when diffing, via `--show-secrets`. Default to off of course


luisdavim (approved):
LGTM, Thanks.

PR #914, opened 2024-07-11
fix segfault on gcp metadata server check

### LINKED ISSUE #909: segfault when using gcs repository
I have a state file that worked fine until I added a `gs://` repository. Now when I run apply, I get a segfault;

```
2024-07-04 14:48:01 INFO: Setting up helm
2024-07-04 14:48:01 INFO: Listing helm repositories
2024-07-04 14:48:01 INFO: Checking Helm version
2024-07-04 14:48:01 INFO: Validating that [ gcs ] is installed
2024-07-04 14:48:02 INFO: Cleaning up sensitive and temp files
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=0xacfeec]

goroutine 1 [running]:
github.com/Praqma/helmsman/internal/gcs.IsRunningInGCP()
        github.com/Praqma/helmsman/internal/gcs/gcs.go:21 +0x2c
github.com/Praqma/helmsman/internal/gcs.Auth()
        github.com/Praqma/helmsman/internal/gcs/gcs.go:47 +0x128
github.com/Praqma/helmsman/internal/app.addHelmRepos(0xc0007265d0)
        github.com/Praqma/helmsman/internal/app/helm_helpers.go:211 +0x76a
        github.com/Praqma/helmsman/internal/app/main.go:76 +0x2f6
main.main()
        github.com/Praqma/helmsman/cmd/helmsman/main.go:9 +0x19
```

This points to this function - and specifically the `resp.Body.Close()` line;

```
func IsRunningInGCP() bool {
	resp, err := http.Get("http://metadata.google.internal")
	resp.Body.Close()
	return err == nil
}
```

My suspicion is that the `resp.Body` is actually `nil`, since the URL cannot be resolved. From the go http docs;

> When err is nil, resp always contains a non-nil resp.Body. Caller should close resp.Body when done reading from it. 

From that I read that if `err` _is not_ `nil`, the `resp.Body` may be `nil`.

ksawarzynski:
should fix #909

### DIFF
--- a/internal/gcs/gcs.go
+++ b/internal/gcs/gcs.go
@@ -18,7 +18,11 @@ var style aurora.Aurora
 
 func IsRunningInGCP() bool {
 	resp, err := http.Get("http://metadata.google.internal")
-	resp.Body.Close()
+
+	if resp != nil && resp.Body != nil {
+		defer resp.Body.Close()
+	}
+
 	return err == nil
 }
 


internal/gcs/gcs.go
```diff
@@ -18,8 +18,14 @@ var style aurora.Aurora
...
-	resp.Body.Close()
-	return err == nil
+	
+	if err != nil {
+		return false
+	}
+
+	defer resp.Body.Close()
+
+	return true
```
thomaspoignant, line 28: I would suggest to avoid the return `true`, and return `false` pattern
```suggestion
	if resp != nil && resp.Body != nil {
		defer resp.Body.Close()
	}
	return err == nil
```

ksawarzynski: done

internal/gcs/gcs.go
```diff
@@ -18,7 +18,11 @@ var style aurora.Aurora
 
 func IsRunningInGCP() bool {
 	resp, err := http.Get("http://metadata.google.internal")
-	resp.Body.Close()
+
+	if resp != nil && resp.Body != nil {
+		defer resp.Body.Close()
```
luisdavim, line 23: Doesn't need to be deferred, the body is not going to be used anymore.

internal/gcs/gcs.go
```diff
@@ -18,7 +18,11 @@ var style aurora.Aurora
 
 func IsRunningInGCP() bool {
 	resp, err := http.Get("http://metadata.google.internal")
-	resp.Body.Close()
+
+	if resp != nil && resp.Body != nil {
```
luisdavim, line 22: This is ok but I don't think we need to check if the body is nil, only the response.

luisdavim: Yes but according to the documentation it won't be nil.

luisdavim (approved):

PR #940, opened 2025-02-16
Fix of urls used in installation section

alsig:
The urls to the releases used in the installation is wrong -> fixed


luisdavim (approved):
Thanks

PR #945, opened 2025-03-25
Fix failing pipelines and update codebase to go 1.24

mkubaczyk:
This should address failing builds from recent days - some packages required golang >= 1.23. I went a bit further and decided it's worth updating straight to 1.24.x and with that I also updated all the dependencies. 

Download url for kubectl changed, I could not get version 1.32.x from the previous one, so I followed the official docs for the curl installation method. 

Even though tests are fine now, I'd appreciate running the build against existing clusters just to confirm everything is fine.


luisdavim (approved):
LGTM, thanks

PR #987, opened 2025-12-19
fix: rely on AWS SDK error propagation for missing credentials

### LINKED ISSUE #975: Prevent lack of AWS_* explicit session variables from generating warnings
This check:
```
func checkCredentialsEnvVar() bool {

	if os.Getenv("AWS_ACCESS_KEY_ID") == "" || os.Getenv("AWS_SECRET_ACCESS_KEY") == "" {

		return false

	} else if os.Getenv("AWS_REGION") == "" {

		if os.Getenv("AWS_DEFAULT_REGION") == "" {
			return false
		}
		os.Setenv("AWS_REGION", os.Getenv("AWS_DEFAULT_REGION"))

	}
	return true
}
```

Results in an excessive amount of WARN logging:
```
2025/10/31 16:50:29 WARN: Failed to find the AWS env vars needed to configure AWS. Please make sure they are set in the environment.
Invalid decryption argument string
2025/10/31 16:50:31 WARN: Failed to find the AWS env vars needed to configure AWS. Please make sure they are set in the environment.
Invalid decryption argument string
2025/10/31 16:50:31 WARN: Failed to find the AWS env vars needed to configure AWS. Please make sure they are set in the environment.
Invalid decryption argument string
```

etc.,

Not having AWS_ACCESS_KEY_ID et al set is a totally valid way to deploy Helmsman, I've been using AWS_PROFILE/KUBECONFIG with Helmsman for years and haven't had any issues, but over the years I have received several messages from engineers in the org wondering if this is related to other Helm/deployment-specific errors that they are seeing. It is a bit mis-leading, and I would expect that just returning the AWS authentication failure back to the user would be more clear in the event that auth isn't properly set

mkubaczyk:
Should fix #975 

I think we can rely on error propagation from AWS SDK instead of upfront requirements check, see https://pkg.go.dev/github.com/aws/aws-sdk-go/aws/credentials#pkg-variables `ErrNoValidProvidersFoundInChain`.
Added simple test to make sure any error is propagated to cover this case.

### DIFF
--- a/internal/aws/aws.go
+++ b/internal/aws/aws.go
@@ -15,27 +15,9 @@ import (
 // colorizer
 var style aurora.Aurora
 
-func checkCredentialsEnvVar() bool {
-	if os.Getenv("AWS_ACCESS_KEY_ID") == "" || os.Getenv("AWS_SECRET_ACCESS_KEY") == "" {
-		return false
-	} else if os.Getenv("AWS_REGION") == "" {
-
-		if os.Getenv("AWS_DEFAULT_REGION") == "" {
-			return false
-		}
-		os.Setenv("AWS_REGION", os.Getenv("AWS_DEFAULT_REGION"))
-
-	}
-	return true
-}
-
 // ReadFile reads a file from S3 bucket and saves it in a desired location.
 func ReadFile(bucketName string, filename string, outFile string, noColors bool) {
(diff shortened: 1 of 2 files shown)


luisdavim (approved):
LGTM

